PR #5, opened 2018-03-10
Adds support for ip whitelisting

fox1t:
First try on whitelist implementation.


fox1t:
I was wondering the same. If you agree I am going to reverse the if statement and fix headers.

delvedor:
Go ahead!

PR #6, opened 2018-04-11
Added redis store support

delvedor:
As titled.
In this way now we support multi-server rate limiting 🎉 

cc @lucamaraschi


index.js
```diff
@@ -15,52 +16,89 @@ const serializeError = FJS({
...
+    .incr(key)
+    .pttl(key)
+    .exec((err, result) => {
+      if (err) return cb(err)
+      if (result[1][1] === -1) {
+        this.store.pexpire(key, this.timeWindow, noop)
+      }
+      cb(null, result[0][1])
+    })
+}
```
mcollina: Can you move those two into separate files?

index.js
```diff
@@ -15,52 +16,89 @@ const serializeError = FJS({
...
     }
   }
 
   next()
 }
 
+function LocalStore (store, timeWindow) {
+  this.store = store
+  const interval = setInterval(this.store.reset.bind(this.store), timeWindow)
+  if (interval.unref) interval.unref()
```
mcollina: why interval might not have `unref()`?

store/RedisStore.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const noop = () => {}
+
+function RedisStore (store, timeWindow) {
+  this.store = store
```
mcollina: this should be `db`.

README.md
```diff
@@ -47,13 +47,16 @@ fastify.register(require('fastify-rate-limit'), {
   max: 3, // default 1000
   timeWindow: 5000, // default 1000 * 60
   cache: 10000, // default 5000
-  whitelist: ['127.0.0.1'] // default []
+  whitelist: ['127.0.0.1'], // default []
+  store: new Redis({ host: '127.0.0.1' }) // default null
```
mcollina: the option should be something like `redis` or `db`.

index.js
```diff
@@ -15,52 +18,57 @@ const serializeError = FJS({
...
+        res.end(serializeError({
+          statusCode: 429,
+          error: 'Too Many Requests',
+          message: `Rate limit exceeded, retry in ${after}`
+        }))
+      }
     }
   }
 
   next()
```
mcollina: Is there any chance that Redis might not be connected? Should we wait until it's ready?

delvedor: I assume that the users passes us a working instance of Redis.

mcollina: a little bit more of safety here would be nice.

store/RedisStore.js
```diff
@@ -0,0 +1,25 @@
...
+  this.key = 'fastify-rate-limit-'
+}
+
+RedisStore.prototype.incr = function (ip, cb) {
+  var key = this.key + ip
+  this.store.pipeline()
+    .incr(key)
+    .pttl(key)
+    .exec((err, result) => {
+      if (err) return cb(err)
```
mcollina: we do not want to bring down the full system if Redis goes down. Maybe we need to handle this error more gracefully? If Redis is down, we let everything go through?

index.js
mcollina: it's probably better to verify. We don't really want to start listening with a broken Redis. Or maybe we want, but that needs to be checked properly.

store/RedisStore.js
delvedor: I don't know, personally I would block everything until redis is reachable again.
I've put a rate limiter for a reason, if I can't guarantee the limits I prefer that everything is blocked instead of leave the api unsafe.

index.js
mcollina: even if it's handled automatically, you truly want your Redis connection to be there when your server starts listening.

delvedor: As far as I know [`ioredis`](https://github.com/luin/ioredis) does not have a way to know if the database is working or not.
We can listen for redis events, but I think it's an overkill solution.
Otherwise we can try a basic query, but I guess it's overkill as well.

I think that if you are using redis in your application you are also using [`fastify-redis`](https://github.com/fastify/fastify-redis) which should handle all this problems for you.

lucamaraschi: I agree with @mcollina that Redis MUST be online and available. Note that this should support either redis or sentinel.

delvedor: I've added a `skipOnError` option, checkout the latest code :)

Issue #7, opened 2018-08-20
Consider not using X-Forwarded-For header for detecting the IP address


mcibique:
I tried to use this rate limiter with fastify server behind the proxy and I found two major problems with the detection of the IP address in following code:

```js
var ip = req.headers['X-Forwarded-For'] || req.connection.remoteAddress
```

1. All keys in `req.headers` are in lowercase so the `req.headers['X-Forwarded-For']` is always undefined even if the proxy sends the header to the upstream. That means, `req.connection.remoteAddress` is always used as IP and it's always the same value (the IP of the proxy). This gives an attacker a possibility to make the entire server unavailable for every user.

2. Even if you fix the casing, there is still another issue: You should never use HTTP header as a cache key unless you trust the source. `X-Forwarded-For` header can be crafted on the client. Imagine your fastify server is publicly available (no proxy in front of it), then following requests from the same client are going to be considered as different:

```
$ curl -v -H "X-Forwarded-For: a" http://example.com/
X-RateLimit-Limit: 5
X-RateLimit-Remaining: 4

$ curl -v -H "X-Forwarded-For: b" http://example.com/
X-RateLimit-Limit: 5
X-RateLimit-Remaining: 4

$ curl -v -H "X-Forwarded-For: c" http://example.com/
X-RateLimit-Limit: 5
X-RateLimit-Remaining: 4
```

With this you can bypass the entire rate limiter, by varying the X-Forwarded-For header for each request. The same happens with the proxy in front of it, because proxies usually leave the original `X-Forwarded-For` header and just append its own IP to the end of the string.

There are few candidates which might suit better for the cache key: e.g. nginx adds [X-Real-IP header](https://nginx.org/en/docs/http/ngx_http_realip_module.html) which is always overridden, or you can use a session ID.

I'd like to propose possible solution: use only `req.connection.remoteAddress` by default and provide a configurable option for cases when devs want to change the IP detection behavior: 
```js
fastify.register(require('fastify-rate-limit'), {
  max: 5,
  timeWindow: '1 minute',
  resolveIpAddress: function (req) {
	return req.headers['x-forwarded-for'];
	// or return req.headers['x-real-ip'];
	// or return req.session.id;
  }
});
```
And then the implementation:
```js
var ip;
if (typeof opts.resolveIpAddress === "function") {
	ip = opts.resolveIpAddress(req);
}
if (!ip) {
	ip = req.connection.remoteAddress;
}
```
[...]

PR #8, opened 2018-09-03
Added opts.keyGenerator and used req.ip for IP address resolving

mcibique:
Sorry, took me a while to get back to the #7 . I have added new option `resolveIpAddress`, updated README and added tests. Let me know if there is anything unclear or can be done better. Thanks.


README.md
```diff
@@ -49,16 +49,28 @@ fastify.register(require('fastify-rate-limit'), {
...
-- `whitelist`: array of string of ips to exlude from rate limiting
+- `whitelist`: array of string of ips to exclude from rate limiting.
 - `redis`: by default this plugins uses an in-memory store, which is fast but if you application works on more than one server it is useless, since the data is store locally.<br>
 You can pass a Redis client here and magically the issue is solved. To achieve the maximum speed, this plugins requires the use of [`ioredis`](https://github.com/luin/ioredis).
-- `skipOnError`: if `true` it will skip errors generated by the storage (eg, redis not reachable)
+- `skipOnError`: if `true` it will skip errors generated by the storage (eg, redis not reachable).
+- `resolveIpAddress`: a function for resolving client's IP address. Use in case your fastify server is behind the proxy. You can check for request header `X-Forwarded-For` or anything else attached to the request by your proxy, but only if you trust the provided value. Check out [#7](https://github.com/fastify/fastify-rate-limit/issues/7) how untrusted value can lead to bypassing the rate limiter. Example usage:
```
mcollina: Can you add that explanation to for that somewhere in the README?

mcibique:
Aaah, I didn't know you have trust proxy implemented. Thanks for pointing it out. I see in the fastify [code](https://github.com/fastify/fastify/blob/master/fastify.js#L80) that if you trust the proxy, `req.ip` is determined from the `X-Forwarded-For` header, otherwise it goes for `req.connection.remoteAddress`. This plugin uses the `req.connection.remoteAddress` directly, and that's ignoring the entire logic in fastify. However, it is easy to fix it by replacing the default fallback: 
```js
const resolveIpAddress = typeof opts.resolveIpAddress === 'function'
    ? opts.resolveIpAddress
    : (req) => req.ip
```

I can update the code in PR and also update the README saying that you can use trustProxy, or if you cannot trust the proxy, you can use own `resolveIpAddress` function to determine the IP.

mcollina:
Note that you can pass `trustProxy: function (addr) { }`, because it's based on https://www.npmjs.com/package/proxy-addr.

mcibique:
I was looking into that function but unfortunately that doesn't solve my problem. `proxyaddr` always get the IP address from 'X-Forwarded-For' header, but I cannot use that header, because my fastify server is behind nginx proxy. Nginx is configured to append its IP to `X-Forwarded-For`, doesn't override it, so the value can be spoofed and cannot be trusted. The remote address of the client is sent from nginx in another header called `X-Real-IP`. That one cannot be spoofed and that's why I'd like to have an option to specify my own logic where to look for the IP address.

Currently I have to do this in my server code to make it working:
```
fastify.addHook('onRequest', function (req, res, next) {
  req.headers['X-Forwarded-For'] = req.headers['x-real-ip']
})
```
... and this is very ugly workaround.

PR #16, opened 2018-12-05
fix(package): update tiny-lru to version 4.0.1

mcollina:
Closes #15


cemremengu:
This will fail unless #14 is merged

Issue #10, opened 2018-11-11
Per Route Rate Limit


csprance:
Is it possible to use this rate-limiter per route instead of for every route?

delvedor:
Currently not, but from Fastify v2 this module will need a refactor and we can think about adding that feature :)

poppinlp:
+1
This is really a necessary feature. 👍

leonidas-120:
Any updates on this?? Its a really useful feature !!

cemremengu:
@mcollina @delvedor I am thinking a list of paths (or pattern matchers?) called `only` can be provided via opts. If the list is `undefined` rate limit will be applied to all routes. What do you think about this approach?

mcollina:
+1, that would work.

matth-c3:
Hello, 

Any update by any chance ? 

Thanks

mcollina:
It seems a very popular feature that needs a proper solution. I'll write my thoughts on this one, as an alternative implementation to https://github.com/fastify/fastify-rate-limit/pull/22.

It is possible to implement this in this way:

```js
app.get('/some/route', {
  config: {
   rateLimit: {
    max: 1000
  }
})
```

How? You can use that config to change the option passed into a route:

```js
app.addHook('onRoute', (routeOptions) => {
  if (routeOptions.rateLimit) {
    // do something!
    // as an example, add a custom hook
    // routeOptions.preHandler = ...
  }
})
```

PR #22, opened 2018-12-14
feat: support rate limit for diff route

poppinlp:
This PR is for https://github.com/fastify/fastify-rate-limit/issues/10
It's not all like what @cemremengu said in issue, but come from that.

The rate limit strategy is:

- limit key is consist with `ip` and `url` not just `ip`
- could specify global limit and time window which will affect all urls as default
- could disable the global limit
- could specify limit and time window for some special urls which will overwrite global setting

This PR is not complete yet. I don't know if you like this. So i send this early PR right now and i feel free for any ideas.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,34 +17,57 @@ const serializeError = FJS({
 })
 
 function rateLimitPlugin (fastify, opts, next) {
-  const timeWindow = typeof opts.timeWindow === 'string'
+  const globalTimeWindow = typeof opts.timeWindow === 'string'
     ? ms(opts.timeWindow)
     : typeof opts.timeWindow === 'number'
       ? opts.timeWindow
       : 1000 * 60
 
   const store = opts.redis
-    ? new RedisStore(opts.redis, timeWindow)
-    : new LocalStore(timeWindow, opts.cache)
+    ? new RedisStore(opts.redis)
+    : new LocalStore(opts.cache)
 
   const keyGenerator = typeof opts.keyGenerator === 'function'
     ? opts.keyGenerator
     : (req) => req.raw.ip
 
   const skipOnError = opts.skipOnError === true
-  const max = opts.max || 1000
   const whitelist = opts.whitelist || []
-  const after = ms(timeWindow, { long: true })
+  const rules = {}
+
+  if (opts.max) {
+    rules.global = {
+      max: opts.max,
+      timeWindow: globalTimeWindow,
+      after: ms(globalTimeWindow, { long: true })
+    }
+  }
+
+  opts.special && opts.special.length && opts.special.forEach(({ url, max, timeWindow }) => {
+    const ruleMax = max || opts.max
+    const ruleTimeWindow = timeWindow || globalTimeWindow
+
+    rules[url] = {
+      max: ruleMax,
+      timeWindow: ruleTimeWindow,
+      after: ms(ruleTimeWindow, { long: true })
+    }
+  })
 
   fastify.addHook('onRequest', onRateLimit)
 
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -17,35 +17,58 @@ const serializeError = FJS({
...
+
+  if (opts.max) {
+    rules.global = {
+      max: opts.max,
+      timeWindow: globalTimeWindow,
+      after: ms(globalTimeWindow, { long: true })
+    }
+  }
+
+  opts.special && opts.special.length && opts.special.forEach(({ url, max, timeWindow }) => {
```
mcollina: Can you please rename this to something else? I feel `special` is too generic.

poppinlp: LOL. I'm not a native speaker and i feel it's a bad name too.
OK, if you think the strategy is accepted, i'll add more in this PR.

store/LocalStore.js
```diff
@@ -2,14 +2,22 @@
 
 const lru = require('tiny-lru')
 
-function LocalStore (timeWindow, cache) {
-  this.lru = lru(cache || 5000)
-  setInterval(this.lru.clear.bind(this.lru), timeWindow).unref()
```
mcollina: This `setInterval` made this module extremely fast. Can you please restore this?
We should not be be starting a timer for each URL, it's very expensive.

mcollina:
Overall I think this PR is moving into the right direction for adding that feature. I also think it needs more work.

poppinlp:
OK. I'll check it out.

index.js
```diff
@@ -17,35 +17,58 @@ const serializeError = FJS({
...
   fastify.addHook('onRequest', onRateLimit)
 
   function onRateLimit (req, res, next) {
-    var key = keyGenerator(req)
-    if (whitelist.indexOf(key) > -1) {
-      next()
-    } else {
-      store.incr(key, onIncr)
+    const key = keyGenerator(req)
+    const rule = rules[req.raw.url] || rules.global
```
delvedor: I think here we should use `reply.context.config.url`, otherwise, dynamic routes can be DDossed.
`reply.context.config.url` yields tot he original URL declaration, so for dynamic URL it will be `/example/:dynamic`.

index.js
```diff
@@ -17,35 +17,58 @@ const serializeError = FJS({
...
+
+    if (whitelist.includes(key) || !rule) {
```
delvedor: Can you use a more specific check? `whitelist.includes(key) === true` for instance.

store/LocalStore.js
delvedor: 100% agree on this point, maybe a global clock could fix the issue?
lìLike the one implemented in [easy-breaker](https://github.com/delvedor/easy-breaker#caveats).

matth-c3:
Hello, 

What is blocking this PR ? 

thanks.

mcollina:
@matth-c3 after the review of @delvedor, @poppinlp did not update the PR anymore. Would you like to send a fresh one with the same content?

matth-c3:
Hello @mcollina , 

I did a version with their modification and some corrections, but it's a middleware, not a plugin anymore. 

On my side it's better to use `fastify.use('path', middleware(rules))` than using `fastify.register(....)` and have a `hook`. 

Not sure that you are interested in such version.

poppinlp:
@mcollina @matth-c3 
Sry I forgot this PR. Is it still needed then I'll update it recently?

matth-c3:
@poppinlp I'm redoing the plugin if you want to add feature it's time :D . See the other PR in progress.

mcollina:
Closing in favor of #34.

PR #31, opened 2019-04-01
Replace @types/redis with @types/ioredis

pc-jedi:
Since we have a dependency to `ioredis` and not `redis` we should
also use the corresponding `@types`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 import * as http from 'http';
 import * as fastify from 'fastify';
-import { RedisClient } from 'redis';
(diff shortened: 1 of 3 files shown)


pc-jedi:
@mcollina Could you create a new version including this fix?

Eomm:
I will take care of it this evening 👍🏼

Eomm:
@mcollina is this change a semver-major?
The tests are changed so I think it is

mcollina:
This is patch, we changed long ago and did not update the types.

PR #33, opened 2019-04-06
updated plugin following best practice from issue#10.

matth-c3:
The plugin has been reviewed, there is a example folder with an example of usage, I did not prepare the test file yet but the example show all the cases possible. 

Let me know if that's suitable for you. 

Here how it works, with `fastify` we register the `plugin` only once with the `default` parameter value that we would apply if this one is missing in the rules ( or just not set ). 

The main options are : 

`max` : set the max fetch that can do a user during a definite period of time;
`timeWindow` : to set the time frame 
`cache` : not use I think
`whitelist` : is an array of ip which will not be restricted
`redis` : connection redis ( from pool for example )
`skipOnError` : skip errors;
`keyGenerator` : to generate a different key to identify the user in the cache. By default it's the ip, but It could also be the username. To adapt based on the infra, in case of reverse proxy for example.
```
  {
    max: 3000, // default 1000 
    //timeWindow: 1000*60, // default 1000 * 60
    //cache: 10000, // default 5000
    //whitelist: ['127.0.0.1'],
    //redis: redis, // default null
    skipOnError: false, // default false
    //keyGenerator: function(req) { /* ... */ }, // default (req) => req.raw.ip
  }
```

example of `get` with all the parameter and the 2 callback : 

onExceeding: when the user is currently fetching/consuming his rate limit
onExceeded : when the user reached his rate limit.

```
fastify.get('/public/sub-rated-1', {
  config : {
    rateLimit : {
      timeWindow: '1 minute',
      whitelist : ['127.0.2.1'],
      onExceeding : function(req) {
        console.log('callback on exceededing ... executed before response to client. req is give as argument')
      },
      onExceeded : function(req) {
        console.log('callback on exceeded ... to black ip in security group for example, req is give as argument')
      }
    }
  }
},(req, reply) => {
  reply.send({ hello: 'from sub-rated-1 ... using default max value ... ' })
})
```

an other example. Here we can see that `max` was not provided, so it will automatically get the default value set in the plugin definition/registration with `fastify`. In this case it will be `3000`

```
fastify.get('/private',{
  config : {
    rateLimit : {
      whitelist : ['127.0.2.1'],
      timeWindow: '1 minute'
    }
  }
},(req, reply) => {
  reply.send({ hello: 'from ... private' })
})
```

You will notice that whitelist is in the `main config` of the plugin and also in the `rule`. There is actually 2 different scope. 

One scope global : define in the `main conf`. Will be use if you trust the source to access everything without limit;
One scope endpoint : define in the individual `rule`, to allow a source to access only specific endpoint.

Let me know what you think about this. 

Thanks.


index.js
```diff
@@ -1,5 +1,3 @@
-'use strict'
```
Eomm: This is necessary

matth-c3: Hello @Eomm ,
Done, I also adapted the previous test ...

store/LocalStore.js
```diff
@@ -2,14 +2,23 @@
...
+LocalStore.prototype.incr = function (prefix, key, timeWindow, cb) {
+  let keyName = `${prefix}-${key}`
+  let current = this.lru.get(keyName) || 0
+  this.lru.set(keyName, ++current)
+
+  if (!this.timers[keyName]) {
+    this.timers[keyName] = setTimeout(() => {
+      this.lru.delete(keyName)
+      this.timers[keyName] = null
+    }, timeWindow)
```
mcollina: this would need an unref().

The previous approach was way more performant. This is going to create a significant higher amount of timers.. and those are expensive. Could this be refactored to avoid this and keep a global timer? I think keeping a plugin-wide timewindow make sense, otherwise things would become a bit more weird to debug.

I’m also ok on a per-route timewindow.

matth-c3: Hello @mcollina ,

Ok .  I will change that.

matth-c3: Hello @mcollina ,

actually I'm not sure to understand,  would you mind refactor this part. I added the `unref`

matth-c3:
Hello, 

A new version, I corrected `.unref()` in the LocalStore. I also took the liberty to add a new functionality which is to store and check the whiteList in Redis. 

There a 2 kind of whiteList : 
- global : will affect all the endpoint.
- endpoint : will affect only the targeted endpoint.

The `Global whitelist` is set in the plugin when registering it with `fastify.register(...)`. 
The `Endpoint whitelist` is set on the endpoint directly with the `{ config : rateLimit : whitelist : [] }`object. 

To use it, you need to set the option `whiteListInRedis` true when initialising the plugin.

It's also possible to use redis to store the rateLimit, but still to use local array to check the whitelist;

@mcollina regarding Local cache and your comment, feel free to change it the way you want.

store/LocalStore.js
matth-c3: @mcollina 
> This is very good work!
> 
> I would also like to retain the previous behavior. By default, the rate limit is applied to all routes. It can be disabled, and then it’s per-route. We should also support keeping it enabled for all routes, but off or with different settings for one.

done. I also updated the doc with what i've done today. 

please validate.

README.md
```diff
@@ -76,6 +82,91 @@ fastify.register(require('fastify-rate-limit'), {
...
+  reply.send({ hello: 'from ... public' })
+})
+
+fastify.get('/public/sub-rated-1', {
+  config: {
+    rateLimit: {
+      timeWindow: '1 minute',
+      whitelist: ['127.0.0.1'],
+      onExceeding: function (req) {
+        console.log('callback on exceededing ... executed before response to client. req is give as argument')
```
mcollina:
```suggestion
        console.log('callback on exceededing ... executed before response to client')
```

README.md
```diff
@@ -76,6 +82,91 @@ fastify.register(require('fastify-rate-limit'), {
...
+fastify.get('/public', (req, reply) => {
+  reply.send({ hello: 'from ... public' })
+})
+
+fastify.get('/public/sub-rated-1', {
+  config: {
+    rateLimit: {
+      timeWindow: '1 minute',
+      whitelist: ['127.0.0.1'],
+      onExceeding: function (req) {
```
mcollina: Why did you add include `reply` as well?

index.js
```diff
@@ -16,58 +14,153 @@ const serializeError = FJS({
...
-
-  const skipOnError = opts.skipOnError === true
-  const max = opts.max || 1000
-  const whitelist = opts.whitelist || []
-  const after = ms(timeWindow, { long: true })
+  const prefix = {
+    cache: params.prefixCache ? params.prefixCache.replace(/[-,.;#*]/g, ':') : urlT
+  }
+  if (pluginComponent.isRedis && pluginComponent.whiteListInRedis) {
+    prefix.wl = urlT
```
mcollina: What's `urlT`? What's `wl`? Please use long, descriptive variable names.

index.js
```diff
@@ -16,58 +14,153 @@ const serializeError = FJS({
...
+            next()
+          }
+        })
+      }
     } else {
-      store.incr(key, onIncr)
+      if (pluginComponent.whitelist.global.indexOf(key) > -1 || pluginComponent.whitelist.endpoint[prefix.cache].indexOf(key) > -1) {
+        next()
+      }
     }
```
mcollina: I'm not sure about this logic. This PR has already too much change, and it's getting hard to review. Would you mind removing the new whitelist mechanism and send it as a follow-up PR?

index.js
```diff
@@ -16,58 +14,153 @@ const serializeError = FJS({
...
+
     function onIncr (err, current) {
-      if (err && skipOnError === false) return next(err)
+      if (err && params.skipOnError === false) return next(err)
+
+      if (current <= params.max) {
+        res.header('X-RateLimit-Limit', params.max)
+        res.header('X-RateLimit-Remaining', params.max - current)
+
+        if (typeof params.onExceeding === 'function') {
```
mcollina: I'd prefer to pass in a `nop` implementation and avoid the if check.

index.js
```diff
@@ -16,58 +14,153 @@ const serializeError = FJS({
...
+  fastify.addHook('onRoute', (routeOptions) => {
+    if (globalParams.global) {
+      if (routeOptions.config && routeOptions.config.rateLimit && typeof routeOptions.config.rateLimit === 'object') {
+        buildRouteRate(pluginComponent, makeParams(routeOptions.config.rateLimit), routeOptions, next)
+      } else {
+        buildRouteRate(pluginComponent, globalParams, routeOptions, next)
+      }
+    } else {
+      if (routeOptions.config && routeOptions.config.rateLimit && typeof routeOptions.config.rateLimit === 'object') {
+        buildRouteRate(pluginComponent, makeParams(routeOptions.config.rateLimit), routeOptions, next)
```
mcollina: You should not pass `next` here. It's the outer next, and you have already called it when this happens.

README.md
matth-c3: this is in case the user would like to retrieve the ip or other information from the current request. 

But if it's doing intensive tasks not sure that's really useful, also it should maybe be async/await, what do you think.

Your call.

mcollina: Thanks, this is as good as it is now, maybe in a later PR.

index.js
matth-c3:
```
const urlT = (routeOptions.url === '/') ? 'root' : routeOptions.url.replace(/\//g, ':').slice(1)
```

`urlT` is the url transformed in case the user did not give a `prefixCache`. It will also make sure that if it's `/` ( root ), it will replace it with a string as by default I replace the `/` by `:`. 

`wl` is for the `whiteList`, it's a shortname for the variable. 

When using `redis` the key look like this by default : 

**For the whiteList define in the plugin (`global`)  **
````
${pluginComponent.app}:wlg:${prefix.wl}
````

**For the whiteList define in the `route` using { config : rateLimit : { whitelist : [...] } }**
````
${pluginComponent.app}:wle:${prefix.wl}
````

`pluginComponent.app` is the name of the app define in the `plugin` with the option `appName`. It's useful for clarity in redis.

Let me know if further question.

File: index.js (hunk 7 above)
Comment (matth-c3) [reply index.js]: Sorry I don't know `nop`. I will check, you mean the `if ...` ?

Comment (mcollina) [reply index.js]: `function nop () {}`, stick it at the end of the file.

File: index.js (hunk 8 above)
Comment (matth-c3) [reply index.js]: Yes  I had a doubt about that, I will change that.

File: index.js (hunk 6 above)
Comment (matth-c3) [reply index.js]: ok.

File: index.js (hunk 5 above)
Comment (mcollina) [reply index.js]: Can you use log variables, and add the above explanation as a comment?

File: index.js (hunk 7 above)
Comment (matth-c3) [reply index.js]: ok, I will  check :)

File: index.js (hunk 5 above)
Comment (matth-c3) [reply index.js]: Sure, Let me comment the whole code if you want. it will be easier to rewiew.

File: index.js (hunk 8 above)
Comment (matth-c3) [reply index.js]: Do you think it could work to return a new `plugin` from within the `rate-limit-plugin`, like a plugin generated on the fly ? 

What I mean is that I could do something like :   

```
const fp = require('fastify-plugin')
............................
.... some code ....
............................
return fp(buildRouteRate(fastify, {
       components : pluginComponent
       params : , makeParams(routeOptions.config.rateLimit),
       route : routeOptions
}, next))
``` 

instead of : 
```
buildRouteRate(pluginComponent, makeParams(routeOptions.config.rateLimit), routeOptions)

```

Would it be suitable, or it's not really in the logic of the `fp` package ?

Comment (mcollina) [reply index.js]: It won't work. But I don't understand why you'll need this. Also, why are you passing `next` there.

Comment (matth-c3) [reply index.js]: the `next` in `buildRouteRate` is a mistake on my side, as you said earlier it's not needed .

I will commit the update at lunch, Sorry.

File: index.js (hunk 5 above)
Comment (matth-c3) [reply index.js]: Hello @mcollina , commented all the code. Should be better to read.

File: index.js (hunk 8 above)
Comment (matth-c3) [reply index.js]: updated.

File: index.js (hunk 7 above)
Comment (matth-c3) [reply index.js]: I'm sorry I look into it but I might be stupid I'm not sure how to do what you request with the `nop` function, feel free to modify as you wish it to be.

Comment (mcollina) [reply index.js]: When you validate the settings:

```js
params.onExceeding = params.onExceeding || nop
```

README.md
```diff
@@ -76,6 +82,95 @@ fastify.register(require('fastify-rate-limit'), {
...
+    rateLimit: {
+      timeWindow: '1 minute',
+      whitelist: ['127.0.0.1'],
+<<<<<<< HEAD
+      onExceeding: function (req) {},
+=======
+      onExceeding: function (req) {
+        console.log('callback on exceededing ... executed before response to client')
+      },
+>>>>>>> 6c3daf23c35af4d1d62ff685699f2e466386424a
```
mcollina: This should be fixed.

index.js
```diff
@@ -16,59 +14,279 @@ const serializeError = FJS({
...
+
+  globalParams.keyGenerator = typeof settings.keyGenerator === 'function'
+    ? settings.keyGenerator
     : (req) => req.raw.ip
 
-  const skipOnError = opts.skipOnError === true
-  const max = opts.max || 1000
-  const whitelist = opts.whitelist || []
-  const after = ms(timeWindow, { long: true })
+  // const makeParams = (p) => { return { ...globalParams, ...p } }
```
mcollina: Why is this ommented out?

test.js
```diff
@@ -291,3 +320,110 @@ test('With keyGenerator', t => {
...
+//     reply.send('hello!')
+//   })
+//
+//   fastify.inject('/', (err, res) => {
+//     t.error(err)
+//     t.strictEqual(res.statusCode, 200)
+//     t.strictEqual(res.headers['x-ratelimit-limit'], undefined)
+//     t.strictEqual(res.headers['x-ratelimit-remaining'], undefined)
+//   })
+// })
```
mcollina: Why is this commented out?

mcollina:
I'd still prefer to have the code for the new whitelist in its own PR. This is already big as a contribution.

Also, I think the `timeWindow` should be plugin-wise and not configurable in an individual route. I think It should be for each given ip, not per route/ip combination.

index.js
matth-c3: It was crashing during the test on a older version of node.

README.md
matth-c3: Yes sorry .. problem during the merge ...

test.js
matth-c3: This test was not going throw, I prefer us to review the code first if that's fine with you as all 99 previous test goes throw. 
New test need to be created anyway, I will fix this one when the time will come.

matth-c3:
> I'd still prefer to have the code for the new whitelist in its own PR. This is already big as a contribution.
> 

Well from my point of view it's more like a refresh / renew than a contribution.  
No sorry I will not submit a different PR, I spend enough time this morning to comment everything in the clearest way possible. It's the 3rd PR that I do to fastify ( one on the initial plugin review which was not good I can admit it, one on the main core of fastify, and this). 

I'm not greedy or lazy but the code is clear enough, same as the comment. If it's not just let me know and I will simply start a different plugin on my own. Not that I want to be greedy but  

Please be aware that the whitelist in redis is just a start, I already have a working version where params that are adapting/updating dynamically from redis.

> Also, I think the `timeWindow` should be plugin-wise and not configurable in an individual route. I think It should be for each given ip, not per route/ip combination.

Sure as mention already feel free to change it and I will pull your changes later on today.  

Thanks.

jsumners:
> No sorry I will not submit a different PR, I spend enough time this morning to comment everything in the clearest way possible. It's the 3rd PR that I do to fastify ( one on the initial plugin review which was not good I can admit it, one on the main core of fastify, and this).

This is not the right attitude to take when submitting a PR. It is easiest to get changes accepted when 1. the PR is limited to _one_ overall change and 2. you listen to the reviewers's feedback. This PR started out as one thing and then half-way through it you added in more:

https://github.com/fastify/fastify-rate-limit/pull/33#issuecomment-480646571
> A new version, I corrected .unref() in the LocalStore. I also took the liberty to add a new functionality which is to store and check the whiteList in Redis.

This is after your work on the original intent of the PR was commended:

https://github.com/fastify/fastify-rate-limit/pull/33#pullrequestreview-223585117
> This is very good work!

The reviewers are simply trying to help you get the PR into a mergeable state. Big features can take a few days of review back-and-forth to get right. But that whole back-and-forth process is how good, maintainable, code gets included in an OSS project.

matth-c3:
@jsumners in that case take the time that you ( reviewers ) needs to take to do instead of dictating on what I should PR or not. If you need a week to valid those changes than take the week, even two or three but don't comment the way you (yes you) do because it's not helping. 

So far I complied and did what I was asked to do/change/redo ( maybe even more ). If you ( reviewers ) want something to be change, you can do it and we (contributors) would be grateful too as it's how collaborative team work and things are moving.

Again, if you are not satisfy, you think that it's too much or you that's simply too complicated, just close the PR. 

Like yours, my time is precious too. Think about that all of that next time before you answer to someone.

mcollina:
You are probably saying this to one of most open teams in the land of Node.js: this is a friendly and open community. In fact, none of what we said was disrespectful in any form.

The truth is we care a lot about maintenance. Keeping some code around that has some relevant issues is going to cause problems in the future. Landing a change we are not convinced off would mean that we have to rework that PR, or fix a lot of bugs on it in the future. Once this is landed, you might walk away or stay around - I cannot rely on it. As such, I need to minimize that will take to future us to maintain this.

Note that this is an amazing feature and really good work done so far. However, I would like to land this with the minimum diff, as it's almost a rewrite.

I'll push the changes on the `timewindow` when I get the chance.

mcollina:
@matth-c3 can you enable me to push to your repo? It's blocked and I can't push.

mcollina:
I fix my review comments about `LocalStore`, and I removed the Redis whitelist feature.
I might have removed too much, as the code was complex with a lot of nested ifs.

matth-c3:
Ok, just let me know when you merge your change that I can delete the repo. 

Thanks.

matth-c3:
Hello @mcollina , is there any action remaining on my side ?

mcollina:
I would like to use `onRequest` instead of `preHandler` here, so I did https://github.com/fastify/fastify/pull/1594

matth-c3:
Hello @mcollina, on my side I will not use it, just change what you feel like  :)  

maybe it's better for you to create a  new branch instead of working on the fork ?

matth-c3:
closing in favor of [#34](https://github.com/fastify/fastify-rate-limit/pull/34)

Issue #37, opened 2019-05-31
Cannot find module 'ioredis' during TS compile


fox1t:
Hi, i just updated to 2.1.1 and got this error during TS compilation.
It happens because `ioredis` is in devDependencies and you are using it in index.d.ts, that is just a typings file.
Let me know if you need help.

mcollina:
Would you mind sending a pr to remove `ioredis` from `index.d.ts`?

fox1t:
What about using [@types/ioredis](https://www.npmjs.com/package/@types/ioredis) as dependency?

mcollina:
I'm -1 in adding any `@types/ioredis` as dependencies. They increase the download size for everybody, and they are not often greatly maintained.

fox1t:
Then we must use `redis: any`. I'll open a PR.

PR #38, opened 2019-05-31
Removes ioredis as dependency from typings

fox1t:
Fixes issue on importing library in TS project and not installing `@types/ioredis`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,5 @@
 import * as http from 'http';
 import * as fastify from 'fastify';
-import * as ioredis from 'ioredis';
 
 declare namespace fastifyRateLimit {
   interface FastifyRateLimitOptions {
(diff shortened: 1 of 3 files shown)


fox1t:
Fixes https://github.com/fastify/fastify-rate-limit/issues/37

Eomm:
@fox1t could we use the [`/// <reference types="..." />`](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) adding a disclaimer in readme: 

> you may need to install `@types/ioredis`

I'm learning some `typings`, so I'm not 100% sure

Eomm:
> force user installing ioredis types since redis is optional in this case

I agree with this point: I proposed it since it is the same pattern used in fastify for the @type/node dependence but I didn't know all the sides effect, I thought that was necessary only if the property was used in the source code.

So `any` will be ok ✌

fox1t:
Good point on @types/node, but TS on Node.js is unusable without them.  For example,  you can’t even use process.env since process is not defined without them. I consider it a sweet exception, including it as dependency doesn’t hurt the end user.

PR #44, opened 2019-07-30
Custom Error Message (resolves issue #26)

saikatharryc:
#### Changes : Custom Error Message
To send custom error message for all, 
You can pass `errorMessage` during the plugin registration:
```
fastify.register(require('fastify-rate-limit'), {
  global : false, // default true
  max: 3, // default 1000
  timeWindow: 5000, // default 1000 * 60
  cache: 10000, // default 5000
  whitelist: ['127.0.0.1'], // default []
  redis: new Redis({ host: '127.0.0.1' }), // default null
  skipOnError: true, // default false
  keyGenerator: function(req) { /* ... */ }, // default (req) => req.raw.ip
  errorMessage: 'This is your custom error message' //default 'Rate limit exceeded, retry in 1 minute'
})
```

for each route, you can override custom message for each, by supplying `errorMessage` like below:
```
fastify.get('/custom', {
  config: {
    rateLimit: {
      max: 3,
      timeWindow: '1 minute',
      errorMessage: 'Please try after sometime.'
    }
  }
},(req, reply) => {
    reply.send({ hello: 'Hello World!' })
})
```

PR #45, opened 2019-08-04
Add functionality to handle custom error responses

### LINKED ISSUE #26: Custom response
Currently you will always get a JSON in the following schema if you exceed the rate limit:
```json
{
  "error": "Too Many Requests",
  "message": "Rate limit exceeded, retry in 1 minute",
  "statusCode": 429
}
```

Since I use a different JSON schema in my response, it would be great to have a way of customizing the response. E.g. Having a callback to format/send the response.

pc-jedi:
Introduced new parameter `errorMessage` which is a callback to build the custom error response. It has currently to parameters `after` and `max`. It might be also good to provided `req`, but I want to first have simple solution.
If the callback is not provided it defaults to:
```js
function (after, max) {
  return { statusCode: 429, error: 'Too Many Requests', message: `Rate limit exceeded, retry in ${after}` }
}
```

To make this work I had to remove the serialization. Not sure if it would be better to let the user provided this also als a parameter.

Fixes #26 
Closes #44


index.js
```diff
@@ -128,16 +123,11 @@ function buildRouteRate (pluginComponent, params, routeOptions) {
           params.onExceeded(req)
         }
 
-        res.type('application/json').serializer(serializeError)
```
Eomm: I think we could keep this line adding a bool to know if the `errorMessage` has been overwritten by user or not.
This should restore also the changes in tests `t.strictEqual(res.headers['content-type'], 'application/json; charset=utf-8')`

README.md
```diff
@@ -82,6 +85,20 @@ fastify.register(require('fastify-rate-limit'), {
 })
 ```
 
+`errorMessage` example usage:
+```js
+fastify.register(require('fastify-rate-limit'), {
+  /* ... */
+  errorMessage: function(after, max) {
```
delvedor: I would directly add `req` as a first parameter, or change the approach a single object, such as `context`:
```js
{
  after: String,
  max: Number
}
```
The second approach is more future proof :)

pc-jedi: `req` does not contain the `after` or `max`, but might be also needed/nice to have for building error messages.
Is it then better to have to parameter `(req, context)` or shall `req` be part of `context`?
```ts
{
  req: fastify.FastifyRequest<http.IncomingMessage>,
  after: string,
  max: number
}
```

pc-jedi: @delvedor ping :-)

allevo:
I don't like `errorMessage`: the value is a function not a message or string. What do you think?

pc-jedi:
@allevo You're right, me neither. What would be a good alternative?
I want to make clear that this function is only been called if the limit is reached (429).
Then the return value of the function is been used as response body.

allevo:
Maybe `getResponseBody`? But I'm not good in naming...

Eomm:
What about errorMessageFactory?

pc-jedi:
I checked with [fastify docs](https://www.fastify.io/docs/v1.13.x/Server/#seterrorhandler).
There it is `errorHandler`.
But I think `errorResponseFactory` is also a good option.

Eomm:
> I checked with [fastify docs](https://www.fastify.io/docs/v1.13.x/Server/#seterrorhandler).
> There it is `errorHandler`.
> But I think `errorResponseFactory` is also a good option.

Since you are not managing the error I think errorHandler could be confusing (the user may expect to ignore some error for example, but it is mandatory to return a string).

So I would choose between:

- `errorMessage` but it could be a static string or a function (with a complete params signature as suggeste by delvedor)
- `errorWhateverFactory/Builder` in order to explain intuitively that it stands for

README.md
```diff
@@ -56,6 +57,7 @@ fastify.register(require('fastify-rate-limit'), {
   redis: new Redis({ host: '127.0.0.1' }), // default null
   skipOnError: true, // default false
   keyGenerator: function(req) { /* ... */ }, // default (req) => req.raw.ip
+  errorMessage: function(after, max) { /* ... */}, // default (after) => ({ /* ... */ })
```
Eomm:
```suggestion
  errorResponseBuilder: function(req, context) { /* ... */}, 
```

PR #51, opened 2019-09-10
Fixed issue with existing route config

gendronb:
Most recent release(s) of this plugin breaks my existing Fastify projects that are using per-route configuration (used for another plugin I made)...

I'm submitting this quick fix (and additional test) to solve this issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,19 +54,18 @@ function rateLimitPlugin (fastify, settings, next) {
 
   // onRoute add the preHandler rate-limit function if needed
   fastify.addHook('onRoute', (routeOptions) => {
-    if (routeOptions.config) {
-      if (routeOptions.config.rateLimit) {
-        if (typeof routeOptions.config.rateLimit === 'object') {
-          const current = Object.create(pluginComponent)
-          current.store = pluginComponent.store.child(routeOptions)
-          // if the current endpoint have a custom rateLimit configuration ...
-          buildRouteRate(current, makeParams(routeOptions.config.rateLimit), routeOptions)
-        } else if (routeOptions.config.rateLimit === false) {
-          // don't apply any rate-limit
-        } else {
-          throw new Error('Unknown value for route rate-limit configuration')
-        }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -55,15 +55,17 @@ function rateLimitPlugin (fastify, settings, next) {
...
-      } else {
-        throw new Error('Unknown value for route rate-limit configuration')
+      if (routeOptions.config.rateLimit) {
+        if (typeof routeOptions.config.rateLimit === 'object') {
+          const current = Object.create(pluginComponent)
+          current.store = pluginComponent.store.child(routeOptions)
+          // if the current endpoint have a custom rateLimit configuration ...
+          buildRouteRate(current, makeParams(routeOptions.config.rateLimit), routeOptions)
+        } else if (routeOptions.config.rateLimit === false) {
+          // don't apply any rate-limit
```
mcollina: I don't think this condition could actually be triggered, because `routeOptions.config.rateLimit` is already going to be truthy. Would you mind removing it?

mcollina:
Good spot, thanks!

index.js
gendronb: Hi @mcollina you're right, in fact, my fix solves some issues, but not all. The problem is that all tests are made without the assumption of existing per-route configuration for other plugins.

For example, if I add a route config to the other tests, most of them will fail even with my fix.

I will submit an updated PR later today that solves all use-cases...

gendronb:
Hi again @mcollina...

I have simplified the config check statement (to a single line), now works with any pre-existing per-route config.

I also updated all per-route tests to include a default per-route config (to insure we don't break or conflict with other plugin configuration)...

index.js
```diff
@@ -65,6 +65,7 @@ function rateLimitPlugin (fastify, settings, next) {
       } else {
         throw new Error('Unknown value for route rate-limit configuration')
       }
+      // }
```
mcollina: Can you remove this line?

gendronb: Done...

mcollina:
Thanks!

PR #57, opened 2019-10-06
add ratelimit-reset

Eomm:
#### Checklist

- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The response will have some additional headers:
 |`x-ratelimit-limit`     | how many request the client can do
 |`x-ratelimit-remaining` | how many request remain to the client in the timewindow
 |`retry-after`           | if the max has been reached, the millisecond the client must wait before perform new requests
-|`x-ratelimit-reset`     | how many seconds must pass before the rate limit resets (**only** with the redis setup)
+|`x-ratelimit-reset`     | how many seconds must pass before the rate limit resets
 
 ### Options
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -51,6 +51,7 @@ The response will have some additional headers:
 |`x-ratelimit-limit`     | how many request the client can do
 |`x-ratelimit-remaining` | how many request remain to the client in the timewindow
 |`retry-after`           | if the max has been reached, the millisecond the client must wait before perform new requests
+|`x-ratelimit-reset`     | how many seconds must pass before the rate limit resets (**only** with the redis setup)
```
mcollina: Is there some limitations on why we can't offer this without Redis?

Eomm: In the local storage, we are setting up an interval that clear the cache every `timeWindow` sec

So, for what I know, there is not the possibility to get how many ms are lacking to the next interval trigger. 
The solution could be to have a sibling interval that counts the seconds and will be reset together with the cache: could fit this case?

mcollina: You'd need to record the time the last interval has fired, and then count how much time is left.

Eomm: It was obvious 🤯

I will add it 👍

Eomm: Done!

Eomm:
Node 6 some flaky test. Rerun and worked

Eomm:
Do you mean to remove the headers in case of 429?

Maybe it is more related to security and other strategies to protect the endpoints, like ban the client when it does too many 429 requests

mcollina:
Yes, but rate-limiting is about endpoint protection, and the less info we provide to an attacker the better.

Eomm:
Clear, but I would implement in this way in a follow-up PRs:

- add an option to turn on/off the headers when the rate-limit is triggered
- add an "abuse" option that will trigger a 403 when too many rate-limit is reached (inspired by the [GitHub-api](https://developer.github.com/v3/#rate-limiting) )

And could we release this feature when at least one of these features will be added as major?

Issue #63, opened 2019-10-27
Feature: abuse usage


Eomm:
## 🚀 Feature Proposal

Add an `abuse` status that will reply 403 Fobidden when the 429 response has been used to much.

## Motivation

This will let you ban users that are calling the API without managing the 429 response.

## Example

```js
const banList = []
app.register(fastifyRateLimit, {
  abuse: 10, // how many 429 request the plugin can send. It is reset when the rate limit is resetted
});
```

Eomm:
Closing due merge

PR #66, opened 2019-11-17
feature\#63 - Add abuse rate limit.

Svjard:
#### Checklist

For https://github.com/fastify/fastify-rate-limit/issues/63, wasn't sure if the headers should be changed when the abuse limit is hit? Welcome feedback on improvements to this PR, love the idea of the abuse limit.


README.md
```diff
@@ -72,6 +73,7 @@ fastify.register(require('fastify-rate-limit'), {
 
 - `global` : indicates if the plugin should apply the rate limit setting to all routes within the encapsulation scope
 - `max`: is the maximum number of requests a single client can perform inside a timeWindow. It can be a sync function with the signature `(req, key) => {}` where `req` is the Fastify request object and `key` is the value generated by the `keyGenerator`. The function **must** return a number.
+- `ban`: is the maximum number of 429 responses to return to a single client before returning 403. When the ban limit is exceeded the context field will have `ban=true` in the errorResponseBuilder
```
Eomm: We should add the info for "distributed application" like:

```
by default this plugins uses an in-memory store, which is fast but if you application works on more than one server it is useless, since the data is store locally.
```

Svjard: Added the extra note at the top of the Options section to provide some detail about that.

README.md
```diff
@@ -72,6 +75,7 @@ fastify.register(require('fastify-rate-limit'), {
...
```
Eomm: I mean that the ban parameter works in a fuzzy way in a distributed environment, like using the localstore.
(The text I posted is copy-pasted from the `redis` param)

Sorry for the misunderstanding

```suggestion
- `ban`: is the maximum number of 429 responses to return to a single client before returning 403. When the ban limit is exceeded the context field will have `ban=true` in the errorResponseBuilder. This parameter is an in-memory counter and could not works properly in a distributed environment.
```

Svjard: Updated accordingly, agree it is a bit fuzzy. Though it is calculated based on a `current` and `max` while in the time window, so if they use LocalStore then it can be meaningless, if they use Redis then at least `current` is stored in a central fashion so the parameter should have meaning then.

Issue #67, opened 2019-11-19
Feature: Ability to pass in a custom Store


Svjard:
## 🚀 Feature Proposal

Be a nice enhancement to add the ability to pass in a Store interface which abides by the current constructs of Local/Redis store, i.e. incr method and callback. Easy way to extend for other storage mechanisms without bloating the existing package. 

## Motivation

There are a lot of memory storage tools out there, as well as traditional ones such as databases, so often adding a new technology into an already established stack can be costly in terms of maintenance, hosting, etc...

## Example

function thestore = function(timeWindow, key) {

}
etc...

fastify.register(require('fastify-rate-limit'), {
  store: require('./my-custom-store')
})

This means a slightly custom call to the custom-store but otherwise should be a very doable feature.

mcollina:
Seems fair! Would you like to send a PR?

Svjard:
I'll work on a PR for it this week unless someone else beats me to it. :+1:

Eomm:
Merged with #68

PR #68, opened 2019-11-26
#67 - Added the ability to customize the store used for tracking rate limits

Svjard:
#### Checklist


example/example-custom.js
```diff
@@ -0,0 +1,104 @@
...
+          trx
+            .raw(`INSERT INTO RateLimits(Source, Count, TTL) VALUES('${key}',1,${d.TTL || ttl}) ON CONFLICT(Source) DO UPDATE SET Count=Count+1`)
+            .then(() => {
+              cb(null, { current: d.Count ? d.Count + 1 : 1, ttl: d.TTL || ttl })
+            })
+            .catch(err => {
+              cb(err)
+            })
+        }
+      })
```
mcollina: Can you add a catch handler?

index.js
```diff
@@ -42,10 +42,15 @@ function rateLimitPlugin (fastify, settings, next) {
     whitelist: globalParams.whitelist
   }
 
-  if (settings.redis) {
-    pluginComponent.store = new RedisStore(settings.redis, 'fastify-rate-limit-', globalParams.timeWindow)
+  if (settings.store) {
+    const Store = settings.store
+    pluginComponent.store = new Store(globalParams.timeWindow, 'fastify-rate-limit-')
```
Eomm: Why this interface as a constructor?
the `fastify-rate-limit-` was an utill for redis.

What do you think about `globalParams` ?
So user can do a lot of (strange 😄) things as an increment doubled for users that are not in the whitelist

README.md
```diff
@@ -77,6 +77,7 @@ fastify.register(require('fastify-rate-limit'), {
 - `whitelist`: array of string of ips to exclude from rate limiting. It can be a sync function with the signature `(req, key) => {}` where `req` is the Fastify request object and `key` is the value generated by the `keyGenerator`. If the function return a truthy value, the request will be excluded from the rate limit.
 - `redis`: by default this plugins uses an in-memory store, which is fast but if you application works on more than one server it is useless, since the data is store locally.<br>
 You can pass a Redis client here and magically the issue is solved. To achieve the maximum speed, this plugins requires the use of [`ioredis`](https://github.com/luin/ioredis)
+- `store`: a custom store to track requests and rates, for an example custom store please see the [`example/`](https://github.com/fastify/fastify-rate-limit/tree/master/example) folder
```
Eomm: Could we have a minimal store-example here?
The `CustomStore` in test is perfect

index.js
Svjard: @Eomm Agree, that is a better call. I adjusted the PR accordingly and made the examples more explicit.

README.md
Svjard: Added in the README an example as well as pointing them to the examples folder.

Issue #71, opened 2019-12-01
LocalStore onClose does not calls done after clearInterval


samaysar:
When using fastify-rate-limit with LocalStore, in a NestFastifyApplication instance:

await app.close() hangs forever as localStore in the onClose hook,does not call done=>

app. addHook('onClose', (done) => {
    clearInterval(this.interval);
    // missing done();
})

mcollina:
Would you like to send a PR?

samaysar:
Hi,
Sure. Let me prepare it... Will send it you by tomorrow.
Thanks for quick response.

Best,

samaysar:
hi Matteo, for your review and release. Thanks.

Eomm:
Released in 3.0.1
Thanks for the contribution 💪

samaysar:
With pleasure.

PR #72, opened 2019-12-02
fix issue #71 on LocalStore by adding Hook's done callback

samaysar:
Please review, related to (#71). Please add for the next release package. Many thanks.

### DIFF
--- a/store/LocalStore.js
+++ b/store/LocalStore.js
@@ -10,6 +10,7 @@ function LocalStore (timeWindow, cache, app) {
 
   app.addHook('onClose', (done) => {
     clearInterval(this.interval)
+    done()
   })
 
   function beat () {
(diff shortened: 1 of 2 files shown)


mcollina:
Would you mind to add a unit test, so we do not regress?

samaysar:
Ok. I will prepare another PR.

Eomm:
You may update this one 👍

samaysar:
Sorry didn't get time yesterday. Will do.
Btw, would you prefer the full callback as per fastify's hooks document?
Which is addHook('... ', (instance, done) =>{... })???
I see it is using avvio's version with single param... If we stick to
avvio, we can remove 'done' param altogether...

What do you think?

Thanks,

Eomm:
You are right, I think the complete version with `(instance, done)` is better 👍

samaysar:
Thanks for feedback. Let me adjust in the pr.

samaysar:
Hi @Eomm / @mcollina , for your review.

Issue #76, opened 2020-01-03
one route with limit


borodadada:
Hello, i try use limit on one route but not work
example

```
fastifyGlobal.register(require('fastify-rate-limit'),
{
    global : false, // default true
    whitelist: [], // default []
})


fastifyGlobal.get('/', {
    config: {
      rateLimit: {
        max: 3,
        timeWindow: '1 minute'
      }
    }
  }, (req, reply) => {
    reply.send({ hello: 'from ... root' })
  })
```
This code is example from git, not work, request always norm.
If i use global limits WORK! But i no need global limits. 
whats a problem?

mcollina:
Would you mind uploading a full example? Can you add a full script to reproduce?

Svjard:
@borodadada Thanks for the bug report.

I see the issue, its with the timeWindow not being processed `ms` on a per route basis (if you set timeWindow = 60000 the above example will work fine for example) and then causing the cache to clear instantly on a 0 interval with the in-memory store. I'm sending a PR in a minute to fix the issue.

JimKnopf182:
still not working, also with type number!?

`  fastify.post('/', {
    config: {
      rateLimit: {
        timeWindow: 60000,
        max: 6
      }
    },
    schema: { querystring: querystringSchema, params: paramsSchema },
    preHandler: [authHandler],
    handler: handler
  });`

JimKnopf182:
ahh the problem is the authHandler, but I want to limit rate for this handler!? thanks in advance.

mcollina:
I think that's a different bug. Specifically, we need to move the code to use a onRequest hook instead of a preHandler.

Eomm:
This has been fixed on #97 👍

PR #77, opened 2020-01-07
#76 - Added fix to format timewindow on per route config

Svjard:
#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,14 +16,6 @@ const serializeError = FJS({
   }
 })
 
-function formatTimeWindow (tw) {
-  return typeof tw === 'string'
-    ? ms(tw)
-    : typeof tw === 'number'
-      ? tw
-      : 1000 * 60
-}
-
 function rateLimitPlugin (fastify, settings, next) {
   // create the object that will hold the "main" settings that can be shared during the build
   // 'global' will define, if the rate limit should be apply by default on all route. default : true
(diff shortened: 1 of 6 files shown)


Svjard:
README is correct as is, supports both strings and numbers for the time window passed in.

test/route-rate-limit.test.js
```diff
@@ -780,6 +780,39 @@ test('global timeWindow when not set in routes', t => {
   })
 })
 
+test('timeWindow specified as a string', t => {
```
Eomm: The test is passing also without the change, we need to use a custom store in order to test this issue.

index.js
```diff
@@ -84,6 +88,7 @@ function rateLimitPlugin (fastify, settings, next) {
           // load the global timewindow if it is missing from the route config
           routeOptions.config.rateLimit.timeWindow = mergedRateLimitParams.timeWindow
         }
+        routeOptions.config.rateLimit.timeWindow = formatTimeWindow(routeOptions.config.rateLimit.timeWindow)
```
Eomm: This is not necessary because the conversion from string to number has been done in the `makeParams`, check the `mergedRateLimitParams` var.

So we could just add:
```
.. else {
  routeOptions.config.rateLimit.timeWindow = mergedRateLimitParams.timeWindow
}
```

test/route-rate-limit.test.js
Svjard: Alright yep you are right, fails when running standalone, but as a test passes regardless, will change to use a custom store and look at `x-ratelimit-reset` to verify the timeWindow is has been converted correctly.

index.js
Svjard: You are right. So what do you think about this instead (remove that line and the entire if statement before it)
```
const mergedRateLimitParams = makeParams(routeOptions.config.rateLimit)
mergedRateLimitParams.routeInfo = routeOptions
current.store = pluginComponent.store.child(mergedRateLimitParams)
```

I think having ```mergedRateLimitParams``` look like the params that are passed to register the plugin is the intuitive approach but because we still need to do per route configuration, we pass in the ```routeInfo``` so we have method, url, path, etc... (including the entire config of the route). This also does guarantee the timeWindow provided as a number as you said but ```routeInfo.config.rateLimit.timeWindow``` is available in its original form just in case.

So like in RedisStore for example, that method then looks like:
```
  return child
}
```
No longer requires going into ```config.rateLimit.---``` but just expect the parameters at the top level of the object, just like the constructor does.

Thoughts?

Issue #84, opened 2020-03-26
skipOnError does nothing


JimKnopf182:
Hi everybody,

just wanted to know how/what the skipOnError flag is working. I expected that if redis connection is lost I can skip the rate limit with this flag. Is this wrong? But the behaviour is that my endpoint needs a lot of time to answer, when connection is lost to redis.

Thanks in advance.
Markus

mcollina:
That flag is thrown when an error _happens_. If Redis take a look long time to respond or it is reconnecting. You might want to set https://github.com/luin/ioredis#offline-queue as well.

Let us know if this helps.

Eomm:
I strongly suggest checking all the options in `ioredis` since the default values are not good for the rate limit feature

I use these as start setup:
```
{
      "connectionName": "my-connection-name",
      "host": "localhost",
      "port": 6379,
      "db": 1,
      "connectTimeout": 500,
      "maxRetriesPerRequest": 1
    }
```

mcollina:
@Eomm I think we should cover those in the README.

JimKnopf182:
Thanks guys! This is really strange, it seems like when I configure 'enableOfflineQueue': false, that fastify is responding directly with skipping rate limiting. Without this property the response needs the time of connectTimeout.

mcollina:
It's likely that your redis connection is not working

JimKnopf182:
Yea the question is, what does the rate limiting when redis is for some minutes/hours offline? With offlinequeuing in redis connection my memory would become large? Is there just any possibility to just passthrough the requests when redis is offline and let the requests through without rate checking?

mcollina:
> Yea the question is, what does the rate limiting when redis is for some minutes/hours offline?

`skipOnError` enables you to control this. If it's true, rate limiting will be skipped.

> With offlinequeuing in redis connection my memory would become large?

I won't expect memory to grow _too much_ or to crash your process. 

> Is there just any possibility to just passthrough the requests when redis is offline and let the requests through without rate checking?

Likely, if you would like to contribute a new option to do so.

PR #85, opened 2020-03-27
add ioredis notice

Eomm:
Improve by #84


jsumners:
Why not add your full example configuration as you did in the issue?

mcollina:
I agree. It seems CI is failing  :(

Eomm:
> Why not add your full example configuration as you did in the issue?

updated the example 👍

PR #122, opened 2020-12-31
fix error handler call

### LINKED ISSUE #121: Return an error object instead of a POJO.
Currently, we are returning a POJO (and using a custom serializer) instead of an error object, which means that the `.setErrorHandler` API won't be called.
This should be considered a bug, as users can't override the default response nor perform custom actions when a 429 error happens.

Eomm:
Fix #121 

Just a question in the code since it seems to me that a piece of code is no more necessary since the `Error` is managed internally by fastify (for this reason the `application/json` string has changed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ fastify.listen(3000, err => {
 })
 ```
 
-In case a client reaches the maximum number of allowed requests, an error will be send to the user with the status code setted to `429`:
+In case a client reaches the maximum number of allowed requests, an error will be sent to the user with the status code set to `429`:
 ```js
 {
   statusCode: 429,
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ fastify.listen(3000, err => {
 })
 ```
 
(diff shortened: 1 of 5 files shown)


mcollina:
Can you update the docs? I think this is a semver-major change.

delvedor (approved):
LGTM with @mcollina nits.

delvedor (approved):

mcollina (approved):
lgtm

mcollina (approved):
lgtm

README.md
```diff
@@ -33,15 +33,24 @@ fastify.listen(3000, err => {
 })
 ```
 
-In case a client reaches the maximum number of allowed requests, a standard Fastify error will be returned to the user with the status code setted to `429`:
+In case a client reaches the maximum number of allowed requests, an error will be send to the user with the status code setted to `429`:
```
jsumners, line 36:
```suggestion
In case a client reaches the maximum number of allowed requests, an error will be sent to the user with the status code set to `429`:
```

Issue #108, opened 2020-11-03
Follow IETF draft around rate limited


mcollina:
## 🚀 Feature Proposal

Implement https://tools.ietf.org/id/draft-polli-ratelimit-headers-00.html

## Motivation

Allow clients to avoid being throttled out.

mcollina:
Done in v5.3.0

PR #124, opened 2021-02-12
fea(#108): Modified rate limit headers following IETF standard.

inigo-garcia:
#108 

This PR integrates the IETF standard rate-limit format. You can see the  latest version of the document at [draft-ietf-httpapi-ratelimit-headers.md](https://github.com/ietf-wg-httpapi/ratelimit-headers/blob/f6a7bc7560a776ea96d800cf5ed3752d6d397b06/draft-ietf-httpapi-ratelimit-headers.md).

Considering it's still a draft version, it's possible to activate it optionally using the `enableDraftSpec` parameter. 

List of header which may change:
````
x-ratelimit-limit -> ratelimit-limit
x-ratelimit-remaining -> ratelimit-remaining
x-ratelimit-reset -> ratelimit-reset
````

The documentation uses CamelCase format but fastify only supports LowerCase.

#### Checklist

- [x] added `enableDraftSpec` optional parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ fastify.register(require('fastify-rate-limit'), {
   skipOnError: true, // default false
   keyGenerator: function(req) { /* ... */ }, // default (req) => req.raw.ip
   errorResponseBuilder: function(req, context) { /* ... */},
+  enableDraftSpec: true, // default false. Uses IEFT draft header standard 
   addHeaders: { // default show all the response headers when rate limit is reached
     'x-ratelimit-limit': true,
     'x-ratelimit-remaining': true,
@@ -98,6 +99,7 @@ You can pass a Redis client here and magically the issue is solved. To achieve t
 - `keyGenerator`: a function to generate a unique identifier for each incoming request. Defaults to `(req) => req.ip`, the IP is resolved by fastify using `req.connection.remoteAddress` or `req.headers['x-forwarded-for']` if [trustProxy](https://www.fastify.io/docs/master/Server/#trustproxy) option is enabled. Use it if you want to override this behavior
 - `errorResponseBuilder`: a function to generate a custom response object. Defaults to `(req, context) => ({statusCode: 429, error: 'Too Many Requests', message: ``Rate limit exceeded, retry in ${context.after}``})`
 - `addHeaders`: define which headers should be added in the response when the limit is reached. Defaults all the headers will be shown
+- `enableDraftSpec`: if `true` it will change the HTTP rate limit headers following the IEFT draft document. More information at [draft-ietf-httpapi-ratelimit-headers.md](https://github.com/ietf-wg-httpapi/ratelimit-headers/blob/f6a7bc7560a776ea96d800cf5ed3752d6d397b06/draft-ietf-httpapi-ratelimit-headers.md).
 
 `keyGenerator` example usage:
 ```js
@@ -272,6 +274,19 @@ These examples show an overview of the `store` feature and you should take inspi
 
 - [Knex-SQLite](./example/example-knex.js)
 - [Sequelize-PostgreSQL](./example/example-sequelize.js)
+
+
+### IETF Draft Spec Headers
+
+The response will have the following headers if `enableDraftSpec` is `true`:
+
+
+| Header | Description |
+|--------|-------------|
+|`ratelimit-limit`       | how many request the client can do
+|`ratelimit-remaining`   | how many request remain to the client in the timewindow
+|`ratelimit-reset`       | how many seconds must pass before the rate limit resets
+|`retry-after`           | contains the same value in time as `ratelimit-reset`
 
 <a name="license"></a>
 ## License
--- a/index.js
+++ b/index.js
@@ -8,6 +8,20 @@ const RedisStore = require('./store/RedisStore')
 
 const routeRateAdded = Symbol('fastify-rate-limit.routeRateAdded')
 
+let labels = {
+  rateLimit: 'x-ratelimit-limit',
+  rateRemaining: 'x-ratelimit-remaining',
+  rateReset: 'x-ratelimit-reset',
+  retryAfter: 'retry-after'
+}
+
+const draftSpecHeaders = {
+  rateLimit: 'ratelimit-limit',
+  rateRemaining: 'ratelimit-remaining',
+  rateReset: 'ratelimit-reset',
+  retryAfter: 'retry-after'
+}
+
 function rateLimitPlugin (fastify, settings, next) {
   // create the object that will hold the "main" settings that can be shared during the build
   // 'global' will define, if the rate limit should be apply by default on all route. default : true
@@ -15,12 +29,19 @@ function rateLimitPlugin (fastify, settings, next) {
     global: (typeof settings.global === 'boolean') ? settings.global : true
   }
 
+  if (typeof settings.enableDraftSpec === 'boolean' && settings.enableDraftSpec) {
+    globalParams.enableDraftSpec = true
+    labels = draftSpecHeaders
+  }
+
   globalParams.addHeaders = Object.assign({
-    'x-ratelimit-limit': true,
-    'x-ratelimit-remaining': true,
-    'x-ratelimit-reset': true,
-    'retry-after': true
+    [labels.rateLimit]: true,
+    [labels.rateRemaining]: true,
+    [labels.rateReset]: true,
+    [labels.retryAfter]: true
   }, settings.addHeaders)
+
+  globalParams.labels = labels
 
   // define the global maximum of request allowed
   globalParams.max = (typeof settings.max === 'number' || typeof settings.max === 'function')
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by jsumners** [comment]
That is a "draft" document. It is not currently widely accepted as the standard. This change should add a `enableDraftSpec` option to enable this new headers. That would allow this to go out without a major release. A major release that changes to the published spec can happen when that spec is solidified and accepted.

**Comment by inigo-garcia** [comment]
Great idea @jsumners , I will add it.

File: README.md [hunk 1]
```diff
@@ -56,10 +56,10 @@ The response will have some additional headers:
...
 | Header | Description |
 |--------|-------------|
-|`x-ratelimit-limit`     | how many request the client can do
-|`x-ratelimit-remaining` | how many request remain to the client in the timewindow
-|`x-ratelimit-reset`     | how many seconds must pass before the rate limit resets
-|`retry-after`           | if the max has been reached, the millisecond the client must wait before perform new requests
+|`ratelimit-limit`     | how many request the client can do
+|`ratelimit-remaining` | how many request remain to the client in the timewindow
+|`ratelimit-reset`     | how many seconds must pass before the rate limit resets
+|`retry-after`         | if the max has been reached, the millisecond the client must wait before perform new requests
```
jsumners, line 62: This is now incorrect.

README.md
```diff
@@ -76,10 +76,11 @@ fastify.register(require('fastify-rate-limit'), {
...
   keyGenerator: function(req) { /* ... */ }, // default (req) => req.raw.ip
   errorResponseBuilder: function(req, context) { /* ... */},
+  enableDraftSpec: true, // default false. Uses IEFT draft header standard 
   addHeaders: { // default show all the response headers when rate limit is reached
-    'x-ratelimit-limit': true,
-    'x-ratelimit-remaining': true,
-    'x-ratelimit-reset': true,
+    'ratelimit-limit': true,
+    'ratelimit-remaining': true,
+    'ratelimit-reset': true,
```
jsumners, line 83: I believe this should remain unchanged.

README.md
```diff
@@ -98,6 +99,7 @@ You can pass a Redis client here and magically the issue is solved. To achieve t
 - `keyGenerator`: a function to generate a unique identifier for each incoming request. Defaults to `(req) => req.ip`, the IP is resolved by fastify using `req.connection.remoteAddress` or `req.headers['x-forwarded-for']` if [trustProxy](https://www.fastify.io/docs/master/Server/#trustproxy) option is enabled. Use it if you want to override this behavior
 - `errorResponseBuilder`: a function to generate a custom response object. Defaults to `(req, context) => ({statusCode: 429, error: 'Too Many Requests', message: ``Rate limit exceeded, retry in ${context.after}``})`
 - `addHeaders`: define which headers should be added in the response when the limit is reached. Defaults all the headers will be shown
+- `enableDraftSpec`: if `true` it will change the HTTP rate limit headers following the IEFT draft document. More information [here](https://github.com/ietf-wg-httpapi/ratelimit-headers/blob/main/draft-ietf-httpapi-ratelimit-headers.md)
```
jsumners, line 102:
```suggestion
- `enableDraftSpec`: if `true` it will change the HTTP rate limit headers following the IEFT draft document. More information at [draft-ietf-httpapi-ratelimit-headers.md](https://github.com/ietf-wg-httpapi/ratelimit-headers/blob/main/draft-ietf-httpapi-ratelimit-headers.md).
```

Also, should reference a specific commit instead of the head.

inigo-garcia: Added to the README, thanks

inigo-garcia:
Thanks for your feedback, I've included all changes

jsumners (approved):
LGTM

mcollina (approved):
lgtm

index.js
```diff
@@ -8,20 +8,38 @@ const RedisStore = require('./store/RedisStore')
...
+}
+
 function rateLimitPlugin (fastify, settings, next) {
   // create the object that will hold the "main" settings that can be shared during the build
   // 'global' will define, if the rate limit should be apply by default on all route. default : true
   const globalParams = {
     global: (typeof settings.global === 'boolean') ? settings.global : true
   }
 
+  const labels = (typeof settings.enableDraftSpec === 'boolean') ? draftSpecHeaders : defaultHeaders
```
simoneb, line 32: if the setting is explicitly set to false, won't this behave as if it was set to true?

inigo-garcia: Yes, I think you are right . I will include it in the PR.

BTW similar behaviour is happening with the global parameter. https://github.com/fastify/fastify-rate-limit/blob/master/index.js#L15

inigo-garcia:
I've found a use case which I'm not sure if it's working as intended.

Consider a scenario where the rate-limit plugin is active and a client is calling the server using a wrong URL (404 request). I believe in this scenario, the current implementation doesn't track, limit or send rate limit headers in the response. Is it correct? 

To test it, I created this test which runs OK
```Javascript
const response = await fastify.inject({
    method: 'GET',
    url: '/notSupported'
  })
  t.strictEqual(response.statusCode, 404)
  t.strictEqual(response.headers['content-type'], 'application/json; charset=utf-8')
  t.notOk(Object.prototype.hasOwnProperty.call(response.headers, 'x-ratelimit-limit'))
  t.notOk(Object.prototype.hasOwnProperty.call(response.headers, 'x-ratelimit-remaining'))
  t.notOk(Object.prototype.hasOwnProperty.call(response.headers, 'x-ratelimit-reset'))
```

mcollina:
@inigo-garcia that might a bug in fastify. Maybe you could open an issue there creating a simple reproduction?

jsumners (approved):

inigo-garcia:
@mcollina I've created a new issue in fastify to discuss about it: https://github.com/fastify/fastify/issues/2855

inigo-garcia:
Based in @simoneb's feedback and the draft I've included another feature. 

The document says:
`If a response contains both the Retry-After and the RateLimit-Reset fields, the value of RateLimit-Reset SHOULD reference the same point in time as Retry-After`

Therefore if `enableDraftSpec` is enabled and the response includes `resetAfter` then it will have the same value as `ratelimit-reset`.

README.md
```diff
@@ -59,7 +59,7 @@ The response will have some additional headers:
 |`x-ratelimit-limit`     | how many request the client can do
 |`x-ratelimit-remaining` | how many request remain to the client in the timewindow
 |`x-ratelimit-reset`     | how many seconds must pass before the rate limit resets
-|`retry-after`           | if the max has been reached, the millisecond the client must wait before perform new requests
+|`retry-after`         | if the max has been reached, the millisecond the client must wait before perform new requests
```
simoneb, line 62: nit: let's keep the same indentation

README.md
```diff
@@ -273,6 +275,19 @@ These examples show an overview of the `store` feature and you should take inspi
...
+
+The response will have the following headers if `enableDraftSpec` is `true`:
+
+
+| Header | Description |
+|--------|-------------|
+|`ratelimit-limit`     | how many request the client can do
+|`ratelimit-remaining` | how many request remain to the client in the timewindow
+|`ratelimit-reset`     | how many seconds must pass before the rate limit resets
+|`retry-after`           | contains the same value in time as `ratelimit-reset`
```
simoneb, line 289: nit: use same indentation for better formatting

index.js
```diff
@@ -159,10 +182,13 @@ function buildRouteRate (pluginComponent, params, routeOptions) {
...
 
-        if (params.addHeaders['x-ratelimit-limit']) { res.header('x-ratelimit-limit', maximum) }
-        if (params.addHeaders['x-ratelimit-remaining']) { res.header('x-ratelimit-remaining', 0) }
-        if (params.addHeaders['x-ratelimit-reset']) { res.header('x-ratelimit-reset', Math.floor(ttl / 1000)) }
-        if (params.addHeaders['retry-after']) { res.header('retry-after', params.timeWindow) }
+        if (params.addHeaders[params.labels.rateLimit]) { res.header(params.labels.rateLimit, maximum) }
+        if (params.addHeaders[params.labels.rateRemaining]) { res.header(params.labels.rateRemaining, 0) }
+        if (params.addHeaders[params.labels.rateReset]) { res.header(params.labels.rateReset, timeLeft) }
+        if (params.addHeaders[params.labels.retryAfter]) {
+          const resetAfterTime = (typeof params.enableDraftSpec === 'boolean' && params.enableDraftSpec) ? timeLeft : params.timeWindow
```
simoneb, line 189: do we need to check again if it's a boolean here?

simoneb (changes requested):
The docs don't include an explicit reference to the new behavior of retry-after. It's mentioned but can easily go unnoticed.

simoneb:
As for the 404 error (or other errors in general), unless the spec mentioned that those need to be handled we don't need to include it here.

index.js
inigo-garcia: You are right, I removed it

inigo-garcia:
Thanks for your feedback @simoneb  . I have added the requested changes

simoneb (approved):

inigo-garcia:
@mcollina @jsumners @simoneb What do you think? Is it ready for merge? 

Finally It has only been implemented the `MUST` requirements and they are only active with `enableDraftSpec` parameter. Thanks

mcollina:
ready to go for me

Issue #125, opened 2021-02-15
Being able to async "max"


leny32:
## 🚀 Feature Proposal

Being able to use async / await in `max` in so we can fetch details from our database/cache or similar.

## Motivation

I would love to contribute to this, and it's a very easy implementation. I've not tested out the performance that might get lost using async / await here.

## Example

For one of my recent projects, different users with different access were able to have different rate limits using async. I think this would be very good to have added to the fastify-rate-limit plugin.

PR #126, opened 2021-02-15
Using async in "max"

leny32:
Here is async compatibility with `max` from the issue: #125 

#### Checklist

- [ x ] run `npm run test` and `npm run benchmark`
- [ x ] tests and/or benchmarks are included
- [ ? ] documentation is changed or added
- [ x ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)


leny32:
This worked when I last tested it.

index.js
```diff
@@ -67,11 +67,11 @@ function rateLimitPlugin (fastify, settings, next) {
   }
 
   // onRoute add the onRequest rate-limit function if needed
-  fastify.addHook('onRoute', (routeOptions) => {
+  fastify.addHook('onRoute', async (routeOptions) => {
```
mcollina, line 70: Unfortunately using `async` here is not supported.

index.js
```diff
@@ -67,11 +67,11 @@ function rateLimitPlugin (fastify, settings, next) {
...
     if (routeOptions.config && typeof routeOptions.config.rateLimit !== 'undefined') {
       if (typeof routeOptions.config.rateLimit === 'object') {
         const current = Object.create(pluginComponent)
-        const mergedRateLimitParams = makeParams(routeOptions.config.rateLimit)
+        const mergedRateLimitParams = await makeParams(routeOptions.config.rateLimit)
```
mcollina, line 74: why you need to do this asynchronously?

mcollina (changes requested):
Can you please add a unit test?

index.js
leny32: I'll fix this.

leny32:
Has been tested using:
```js
const main = async (fastify) => {
	await fastify.register(require('fastify-rate-limit'), {
		keyGenerator: function (req) { return req.headers['authorization']; },
		max: async (_, key) => {
			let { access } = JSON.parse(await fastify.redis.get('API_KEY_' + key));
			return getAmount(access || 0);
		},
		skipOnError: true,
		errorResponseBuilder: function () {
			return {
				code: 429,
				error: 'Too Many Requests',
				message: 'Hourly limit reached, contact staff for increased limit',
				date: Date.now()
			};
		},
		timeWindow: 60 * 60 * 1000,
		redis: new ioredis({
			host: '127.0.0.1',
			port: 6379,
		})
	});
};
```

mcollina:
Can you transform your snippet in a unit test? Check the `test/` folder.

leny32:
I'm not very familiar with unit testing, but I'll try.

PR #123, opened 2021-01-06
do not set onRequest twice

Saniol:
fixes #120 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #130, opened 2021-02-24
fix: registration of multiple rate limiter

### LINKED ISSUE #129: PR #123 is preventing use cases of multiple rate limiter usage
## 💥 Regression Report

Since PR #123 it is not possible to register two rate limiter like for a use case where a global rate limiter should be applied and a second, stricter limiter to only some specific routes like this (nestjs registration of the rate limiter):

```TS
    // rate limiter for every route except /auth/password
    await app.register(fastifyRateLimit, {
      max: 150,
      timeWindow: 1000 * 60 * 5, // 150 req/5min
      redis: new Redis({ ...configService.redisConnectionData, keyPrefix: 'ratelimit:' }),
      whitelist: (req) => {
        return req.url === '/auth/password';
      },
      errorResponseBuilder: rateLimitErrorResponse,
    });

    // rate limiter for /auth/password route only
    await app.register(fastifyRateLimit, {
      max: 15,
      timeWindow: 1000 * 60 * 5, // 15 req/5min
      redis: new Redis({ ...configService.redisConnectionData, keyPrefix: 'ratelimit:auth:' }),
      whitelist: (req) => {
        return req.url !== '/auth/password';
      },
      errorResponseBuilder: rateLimitErrorResponse,
    });
```

## Last working version

Worked up to version: 5.0.0

Stopped working in version: 5.0.1

## To Reproduce

Use two rate limiters, the second one registered does not work.

## Expected behavior

Multiple rate limiter should be usable for a use case where a stricter rate limiting is needed for some routes.

## Your Environment

- *node version*: 14
- *fastify version*: >3.0.0
- *os*: WSL2

squareloop1:
Closes #129 

Fixes regression of PR #123 (v5.0.1) by moving symbol inside function scope to be able to register multiple rate limiters again.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -118,14 +118,6 @@ async function rateLimitPlugin (fastify, settings) {
 }
 
 async function buildRouteRate (pluginComponent, params, routeOptions) {
-  const routeRateAdded = Symbol('fastify-rate-limit.routeRateAdded')
-
-  if (routeOptions[routeRateAdded]) {
-    return
-  }
-
-  routeOptions[routeRateAdded] = true
-
   const after = ms(params.timeWindow, { long: true })
 
   if (Array.isArray(routeOptions.onRequest)) {
--- a/index.js
+++ b/index.js
@@ -6,8 +6,6 @@ const ms = require('ms')
 const LocalStore = require('./store/LocalStore')
 const RedisStore = require('./store/RedisStore')
 
-const routeRateAdded = Symbol('fastify-rate-limit.routeRateAdded')
-
 let labels = {
   rateLimit: 'x-ratelimit-limit',
   rateRemaining: 'x-ratelimit-remaining',
@@ -120,6 +118,8 @@ async function rateLimitPlugin (fastify, settings) {
 }
 
 async function buildRouteRate (pluginComponent, params, routeOptions) {
+  const routeRateAdded = Symbol('fastify-rate-limit.routeRateAdded')
+
   if (routeOptions[routeRateAdded]) {
     return
   }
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
Can you please add a unit test?

squareloop1:
A test for this should already be included by PR #123 or is this test not sufficient for this case?

I am also asking myself, if this code makes any sense at all, since we are creating the symbol inside the function the first conditional that checks for this symbol is never true and therefore unnecessary, or am I missing something here?

mcollina:
> A test for this should already be included by PR #123 or is this test not sufficient for this case?

No, because if the behavior you are looking for was covered by tests, we would not have regressed.

> I am also asking myself, if this code makes any sense at all, since we are creating the symbol inside the function the first conditional that checks for this symbol is never true and therefore unnecessary, or am I missing something here?

Unfortunately #123 address a different problem as fastify might register the same route _twice_ in certain cases. This is fixed in Fastify already, so maybe that could be reverted.

squareloop1:
> Unfortunately #123 address a different problem as fastify might register the same route twice in certain cases. This is fixed in Fastify already, so maybe that could be reverted.

So maybe we should just revert the changes introduced by #123 instead of merging this PR if it was fixed in fastify already anyways?

mcollina:
Yes, that would be a viable solution. I would still prefer to have a test with a route where this plugin is installed twice to avoid regressing this.

squareloop1:
I removed the changes introduced in #123 but kept the test case from it and added another test case for multiple rate limiter registrations

mcollina (approved):
lgtm

Issue #127, opened 2021-02-18
Rate limit 404s


mcollina:
## 🚀 Feature Proposal

It would be interesting to rate limit 404s as well. This might be controversial because it's faster to actually return a 404 vs read up the current limit from a database. However it might be a security risk as an attacker could search which routes are available.

## Example

```js
  const fastify = Fastify()
  await fastify.register(rateLimit, { global: true, max: 2, timeWindow: 1000 })
  fastify.setNotFoundHandler({
    onRequest: fastify.rateLimit
  }, function (request, reply) {
    reply.code(404).send({ hello: 'world' })
  })
  fastify.get('/', function (request, reply) {
    reply.send({ hello: 'world' })
  })
  await fastify.listen(3000)
```

jsumners:
Sounds good to me.

nherment:
I would appreciate any tips to send a PR on this.

After looking at the fastify docs on error handling an ideal solution is not obvious.

Is it possible with fastify to add a hook around the notFoundHandler and still allow users to customize the handler?

mcollina:
You should expose https://github.com/fastify/fastify-rate-limit/blob/4856e340ceb6d0c22663bc47cf3dced8dd74a3df/index.js#L139 via `fastify.decorate('rateLimit')`, then the snippet above will work.

(We also need documentation)

PR #140, opened 2021-03-10
Add ability to rate limit 404 handler

nherment:
Addresses #127

This adds the ability for users to rate limit the setNotFoundHandler.
An attacker could search for valid URLs if your 404 error handling is not rate limited.

#### Checklist

- [ x] run `npm run test` ~and `npm run benchmark`~ (comment: `benchmark` script does not exist)
- [ x] tests and/or benchmarks are included
- [ x] documentation is changed or added
- [ x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,22 @@ To rate limit your 404 response, you can use a custom handler:
 const fastify = Fastify()
 await fastify.register(rateLimit, { global: true, max: 2, timeWindow: 1000 })
 fastify.setNotFoundHandler({
-  preHandler: fastify.rateLimit
+  preHandler: fastify.rateLimit()
+}, function (request, reply) {
+  reply.code(404).send({ hello: 'world' })
+})
+```
+
+Note that you can customize the behaviour of the preHandler in the same way you would for specific routes:
+
+```js
+const fastify = Fastify()
+await fastify.register(rateLimit, { global: true, max: 2, timeWindow: 1000 })
+fastify.setNotFoundHandler({
+  preHandler: fastify.rateLimit({
+    max: 4,
+    timeWindow: 500
+  })
 }, function (request, reply) {
   reply.code(404).send({ hello: 'world' })
 })
--- a/README.md
+++ b/README.md
@@ -61,6 +61,22 @@ The response will have some additional headers:
 |`x-ratelimit-reset`     | how many seconds must pass before the rate limit resets
 |`retry-after`           | if the max has been reached, the millisecond the client must wait before perform new requests
 
+
+### Preventing guessing of URLS through 404s
+
+An attacker could search for valid URLs if your 404 error handling is not rate limited.
+To rate limit your 404 response, you can use a custom handler:
+
+```js
+const fastify = Fastify()
+await fastify.register(rateLimit, { global: true, max: 2, timeWindow: 1000 })
+fastify.setNotFoundHandler({
+  preHandler: fastify.rateLimit
+}, function (request, reply) {
+  reply.code(404).send({ hello: 'world' })
+})
+```
+
 ### Options
 
 You can pass the following options during the plugin registration:
@@ -76,7 +92,7 @@ fastify.register(require('fastify-rate-limit'), {
   skipOnError: true, // default false
   keyGenerator: function(req) { /* ... */ }, // default (req) => req.raw.ip
   errorResponseBuilder: function(req, context) { /* ... */},
-  enableDraftSpec: true, // default false. Uses IEFT draft header standard 
+  enableDraftSpec: true, // default false. Uses IEFT draft header standard
   addHeaders: { // default show all the response headers when rate limit is reached
     'x-ratelimit-limit': true,
     'x-ratelimit-remaining': true,
@@ -211,7 +227,7 @@ fastify.register(require('fastify-rate-limit'),
   {
     global : false, // don't apply these settings to all the routes of the context
     max: 3000, // default global max rate limit
-    allowList: ['192.168.0.10'], // global allowlist access. 
+    allowList: ['192.168.0.10'], // global allowlist access.
     redis: redis, // custom connection to redis
   })
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -11,7 +11,7 @@ declare module 'fastify' {
     ip: string | number
   }
   interface FastifyInstance {
-    rateLimit: (options?:RateLimitPluginOptions) => preHandlerAsyncHookHandler;
+    rateLimit: (options?:RateLimitOptions) => preHandlerAsyncHookHandler;
   }
 }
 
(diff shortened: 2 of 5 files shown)


climba03003:
linting is failed. Can you fix the linting error?

climba03003:
The `rateLimit` handler can be handle like register a database. Then, it can allow register multiple times.
Example Code:
https://github.com/fastify/fastify-mongodb/blob/master/index.js#L24-L39

index.d.ts
```diff
@@ -10,6 +10,9 @@ declare module 'fastify' {
     > {
     ip: string | number
   }
+  interface FastifyInstance {
+    rateLimit: preHandlerAsyncHookHandler;
+  }
```
nherment, line 15: I haven't tested this but believe it's accurate.

index.js
```diff
@@ -91,6 +91,11 @@ async function rateLimitPlugin (fastify, settings) {
   pluginComponent.run = run
   fastify.decorateRequest(run, false)
 
+  if (!fastify.hasDecorator('rateLimit')) {
+    // The rate limit plugin can be registered multiple times but decorate throws if called multiple times for the same field
+    fastify.decorate('rateLimit', rateLimitRequestHandler(globalParams, pluginComponent))
+  }
```
nherment, line 97: There are 2 things I don't like with this implementation but it seems to be what was suggested in issue #127:
1. The rateLimit uses the global params without the ability to customize that at the `preHandler` level of the `setNotFoundHandler`
2. The plugin can supposedly be registered more than once but this implementation will only retain the first registration to the decorator.

mcollina (changes requested):

README.md
```diff
@@ -61,6 +61,22 @@ The response will have some additional headers:
...
+### Preventing guessing of URLS through 404s
+
+An attacker could search for valid URLs if your 404 error handling is not rate limited.
+To rate limit your 404 response, you can use a custom handler:
+
+```js
+const fastify = Fastify()
+await fastify.register(rateLimit, { global: true, max: 2, timeWindow: 1000 })
+fastify.setNotFoundHandler({
+  preHandler: fastify.rateLimit
```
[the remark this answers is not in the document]
nherment: The documentation differs: https://www.fastify.io/docs/latest/Server/#setnotfoundhandler

```js
fastify.setNotFoundHandler({
  preValidation: (req, reply, done) => {
    // your code
    done()
  },
  preHandler: (req, reply, done) => {
    // your code
    done()
  }
}, function (request, reply) {
    // Default not found handler with preValidation and preHandler hooks
})
```

index.js
nherment: Yes good point. Like you suggested, I made `rateLimit` a function so that it makes the signature more forward compatible.

index.d.ts
nherment: A proper test of the `rateLimit()` decorator used on the `setNotFoundHandler` method will require a fix on fastify to expose the optional hooks to the method. Cf. https://github.com/fastify/fastify/pull/2929

test/not-found-handler-rate-limited.js
```diff
@@ -0,0 +1,64 @@
...
+          t.strictEqual(res.headers['x-ratelimit-remaining'], 0)
+          t.strictEqual(res.headers['retry-after'], 1000)
+          t.strictEqual(res.headers['x-ratelimit-reset'], 0)
+          t.deepEqual(JSON.parse(res.payload), {
+            statusCode: 429,
+            error: 'Too Many Requests',
+            message: 'Rate limit exceeded, retry in 1 second'
+          })
+
+          setTimeout(retry, 1100)
```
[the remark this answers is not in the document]
nherment: Yes I'll remove it altogether. Other tests cover this use case.

README.md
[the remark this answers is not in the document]
mcollina: I'm fairly sure that `onRequest` is supported. Can you verify?

index.d.ts
mcollina: Inside https://github.com/fastify/fastify-rate-limit/blob/master/test/types/types.test.ts just call the function and add an assertion to validate the returned type.

index.js
mcollina: > Yes good point. Like you suggested, I made rateLimit a function so that it makes the signature more forward compatible.

The current code does not match what I recommended.

README.md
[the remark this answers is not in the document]
nherment: `onRequest` works too.

index.js
climba03003: use it as function should allow user to pass in `options` and which they can customize the `rateLimitRequestHandler`. I think it can be extracted to be a big `createRateLimit` function and which you can use it in both `plugin` and `decorators`.

nherment: @mcollina @climba03003 ok so if I understand what you are saying is that you would like the configuration `fastify.rateLimit(opts)` to be part of this PR as opposed to having an empty placeholder `fastify.rateLimit(void)` for now ?

climba03003: Yes, it is.

README.md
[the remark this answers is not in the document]
nherment: @mcollina what do you recommend given it's `preHandler` that is documented?

mcollina (approved):
lgtm

index.d.ts
```diff
@@ -10,6 +10,9 @@ declare module 'fastify' {
     > {
     ip: string | number
   }
+  interface FastifyInstance {
+    rateLimit: (options?:RateLimitPluginOptions) => preHandlerAsyncHookHandler;
```
nherment, line 14: These are the same options as the plugin. From reading all of them it mostly makes sense but please let me know if you would like these options to be different.
For example, the `global` boolean does  not seem necessary.

climba03003 (approved):
LGTM. Great Work.

index.d.ts
climba03003: Then, use `Omit<RateLimitPluginOptions, 'global'>`

nherment:
Currently these options cannot be overridden in the rateLimit() and by the route definitions:
- `global`
- `cache`
- `redis`
- `addHeaders`

It makes sense for the first 3 to be defined at the plugin level but should the `addHeaders` also be available on the routes/`rateLimit()` func?

mcollina:
I would not worry about it right now @nherment

index.d.ts
nherment: I split the TS definitions instead to only expose the used options on `rateLimit(opts)` Hope that's ok.

mcollina (approved):
lgtm

Issue #160, opened 2021-06-06
Broken Example of Custom Store with Knex


58bits:
Fairly sure that the custom store example in /example/example-knex.js  does not work. At a minimum there is a function scope 'this' error, as well as errors in the logic for the inserts and updates. There should also be a read lock on the row before the updates are attempted. The /example-sequelize.js example seems to be okay. The logic is correct, although not sure whether a similar transaction is required.

Below is a working example for Knex.js and MySQL. 

I'd be glad to submit a PR for /example/example-knex-mysql.js

```
'use strict'

// Example of a custom store using Knex.js and MySQL
//
// Note that the rate check should place a read lock on the row.
// For MySQL see:
// https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html
// https://blog.nodeswat.com/concurrency-mysql-and-node-js-a-journey-of-discovery-31281e53572e
//
// Below is an example table to store rate limits that must be created
// in the database first.
// // Knex migration
// exports.up = async knex => {
//   await knex.schema.createTable('rate_limits', table => {
//     table.string('source').notNullable()
//     table.string('route').notNullable()
//     table.integer('count').unsigned()
//     table.bigInteger ('ttl')
//     table.primary(['route', 'source'])
//   })
// }
//
// exports.down = async knex => {
//   await knex.schema.dropTable('rate_limits')
// }
// // Above migration will create...
// CREATE TABLE `rate_limits` (
//   `source` varchar(255) NOT NULL,
//   `route` varchar(255) NOT NULL,
//   `count` int unsigned DEFAULT NULL,
//   `ttl` int unsigned DEFAULT NULL,
//   PRIMARY KEY (`route`,`source`)
// ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;

function KnexStore(options) {
  this.options = options
  this.route = ''
}

KnexStore.prototype.routeKey = function (route) {
  if (route) this.route = route
  return route
}

KnexStore.prototype.incr = function (key, cb) {
  const now = (new Date()).getTime()
  const ttl = now + this.options.timeWindow
  const cond = { route: this.route, source: key }
  knex.transaction(function (trx) {
    trx('rate_limits')
      .whereRaw('route = ? AND source = ? FOR UPDATE', [cond.route || '', cond.source]) // Create read lock
      .then(r => {
        const d = r[0]
        if (d && d.ttl > now) {
          trx
[...]

PR #161, opened 2021-06-07
Updated docs and added example/example-knex-mysql.js

58bits:
Added a working example of Knex.js with MySQL

### DIFF
--- a/example/example-knex-mysql.js
+++ b/example/example-knex-mysql.js
@@ -44,62 +44,61 @@ KnexStore.prototype.routeKey = function (route) {
   return route
 }
 
-KnexStore.prototype.incr = function (key, cb) {
+KnexStore.prototype.incr = async function (key, cb) {
   const now = (new Date()).getTime()
   const ttl = now + this.options.timeWindow
   const max = this.options.max
   const cond = { route: this.route, source: key }
-  knex.transaction(function (trx) {
-    // NOTE: MySQL syntax FOR UPDATE for read lock on counter stats in row
-    trx('rate_limits')
-      .whereRaw('route = ? AND source = ? FOR UPDATE', [cond.route || '', cond.source]) // Create read lock
-      .then(r => {
-        const d = r[0]
+  try {
+    await knex.transaction(async (trx) => {
+      try {
+        // NOTE: MySQL syntax FOR UPDATE for read lock on counter stats in row
+        const row = await trx('rate_limits')
+          .whereRaw('route = ? AND source = ? FOR UPDATE', [cond.route || '', cond.source]) // Create read lock
+        const d = row[0]
         if (d && d.ttl > now) {
           // Optimization - no need to UPDATE if max has been reached.
           if(d.count < max) {
-            trx
-              .raw('UPDATE rate_limits SET count = ? WHERE route = ? AND source = ?', [d.count + 1, cond.route, key])
-              .then(() => {
-                cb(null, { current: d.count + 1, ttl: d.ttl })
-              })
-              .catch(err => {
-                trx.rollback()
-                cb(err, {
-                  current: 0,
-                })
-              })
-          } else { // No need to update the counter if we're already at max. Saves a write to the DB
+            try {
+              await trx
+                .raw('UPDATE rate_limits SET count = ? WHERE route = ? AND source = ?', [d.count + 1, cond.route, key])
+              cb(null, { current: d.count + 1, ttl: d.ttl })
+            } catch(err) {
+              // TODO: Handle as appropriate
+              fastify.log.error(err)
+              cb(err, { current: 0 })
+            }
+          } else { // We're already at max. No UPDATE above saves a write, but we must send d.count + 1 to trigger rate limit.
             cb(null, { current: d.count +1, ttl: d.ttl })
           }
         } else {
-          // NOTE: MySQL syntax for ON DUPLICATE KEY UPDATE
-          trx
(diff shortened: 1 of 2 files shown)


58bits:
A new example custom store for Knex.js and MySQL. 

I didn't feel it was my place to remove the example-knex.js version for SQLite, but I am pretty sure that it is broken in several ways (as per https://github.com/fastify/fastify-rate-limit/issues/160)

Hope this helps.

example/example-knex-mysql.js
```diff
@@ -0,0 +1,147 @@
...
+KnexStore.prototype.incr = function (key, cb) {
+  const now = (new Date()).getTime()
+  const ttl = now + this.options.timeWindow
+  const max = this.options.max
+  const cond = { route: this.route, source: key }
+  knex.transaction(function (trx) {
+    // NOTE: MySQL syntax FOR UPDATE for read lock on counter stats in row
+    trx('rate_limits')
+      .whereRaw('route = ? AND source = ? FOR UPDATE', [cond.route || '', cond.source]) // Create read lock
+      .then(r => {
```
climba03003, line 56: Can you use `async-await` style instead of multiple `Promise` chain inside another `Promise` chain.

climba03003 (changes requested):

example/example-knex-mysql.js
58bits: Sure!

example/example-knex-mysql.js
```diff
@@ -0,0 +1,141 @@
...
+        } catch(err) {
+          // TODO: Handle as appropriate
+          fastify.log.error(err)
+          cb(err, { current: 0 })
+        }
+      }
+    } catch(err) {
+      // TODO: Handle as appropriate
+      fastify.log.error(err)
+      cb(err, { current: 0 })
```
mcollina, line 88:
```suggestion
      process.nextTick(cb, err, { current: 0 })
```

This fixes the error handling of using async/await with a callback.

example/example-knex-mysql.js
```diff
@@ -0,0 +1,141 @@
...
+  if (route) this.route = route
+  return route
+}
+
+KnexStore.prototype.incr = async function (key, cb) {
+  const now = (new Date()).getTime()
+  const ttl = now + this.options.timeWindow
+  const max = this.options.max
+  const cond = { route: this.route, source: key }
+  await knex.transaction(async (trx) => {
```
mcollina, line 52: Can we avoid mixing async/await with a callback?

example/example-knex-mysql.js
```diff
@@ -0,0 +1,141 @@
...
+          cb(err, { current: 0 })
+        }
+      }
+    } catch(err) {
+      // TODO: Handle as appropriate
+      fastify.log.error(err)
+      cb(err, { current: 0 })
+    }
+  })
+}
```
mcollina, line 91: This function should not return a promise

example/example-knex-mysql.js
```diff
@@ -0,0 +1,141 @@
...
+  reply.send({ hello: 'from ... private' })
+})
+
+fastify.get('/public', (req, reply) => {
+  reply.send({ hello: 'from ... public' })
+})
+
+fastify.listen(3000, err => {
+  if (err) throw err
+  console.log('Server listening at http://localhost:3000')
```
mcollina, line 140: remove this

example/example-knex-mysql.js
climba03003: I think yes.
```js
let trx
try {
// Using trx as a transaction object:
  trx = await knex.transaction();
  await trx('') ....
  await trx('') ....
  await trx.commit()
} catch {
  await trx.rollback()
}
```
```

example/example-knex-mysql.js
58bits: Thanks

example/example-knex-mysql.js
58bits: Okay let me try. It may be a few days. Deadline looms. More soon...

example/example-knex-mysql.js
58bits: I've wrapped each callback in process.nextTick as suggested.

example/example-knex-mysql.js
58bits: Thanks for the feedback. I've refactored as suggested.

example/example-knex-mysql.js
58bits: I completely confess to not being entirely sure what this means here.

example/example-knex-mysql.js
58bits: Done

example/example-knex-mysql.js
58bits: Ah - I think I see - since before it was refactored, the effective last statement was the start of the Knex transaction via `await knex.transaction(async (trx) => {...` and so the return value of the resolved transaction would have been returned from `KnexStore.prototype.incr`.

mcollina (approved):
lgtm

58bits:
Thanks Matteo. And thanks to both of you for the very helpful comments and suggestions.

climba03003 (approved):
LGTM.

climba03003:
@58bits 
Can you fix the linting error? I will merge it after you fix those issue.

58bits:
> 
> 
> @58bits
> Can you fix the linting error? I will merge it after you fix those issue.

Done.

PR #179, opened 2021-08-25
Fix/route timewindow is ignored

leorossi:
This is just a draft to understand if the approach is ok.

Basically store's `incr()` function would now require a `timeWindow` parameter, instead of using the one used to create the object.

This because when using a route like this
```js
fastify.route({
    method: 'GET',
    url: '/',
    preHandler: [
      fastify.rateLimit({
        max: 1,
        timeWindow: '2 minutes',
        keyGenerator: function (request) {
          return 45
        }
      })
    ],

    handler: (request, reply) => {
      reply.send({ hello: 'world' })
    }

  })
 ```
the `timeWindow` will be ignored and instead a default value of 60 seconds would be set.

If this approach is accepted I will change the constructor and update all the documentation and tests.

Fixes: #175 

#### Checklist

### CODE REVIEW

**Comment by climba03003** [comment]
As the comment from https://github.com/fastify/fastify-rate-limit/issues/175#issuecomment-905227704

I recommend to use the same logic inside onRoute handler. It should create a child RedisStore if it passed any custom options.
https://github.com/fastify/fastify-rate-limit/blob/master/index.js#L115-L118

The approach you use can solve the problem but change the Store prototype and API structure is a major bump.

**Comment by leorossi** [comment]
@climba03003 problem is that I don't have all the parameters needed for every store implementation inside `rateLimitRequestHandler`. Also those parameters are different from store to store (i.e `RedisStore` has `key`, `LocalStore` has `cache`)

What about a very simple solution like the one below, which will not change the public store's API ?

```
RedisStore.prototype.incr = function(ip, cb) {
  return this.incrementWithCustomTimeWindow(ip, this.timeWindow, cb);
}
RedisStore.prototype.incrementWithCustomTimeWindow = function (ip, timeWindow, cb) { ... }
```

**Comment by climba03003** [comment]
> 
> 
> @climba03003 problem is that I don't have all the parameters needed for every store implementation inside `rateLimitRequestHandler`. Also those parameters are different from store to store (i.e `RedisStore` has `key`, `LocalStore` has `cache`)
> 
> What about a very simple solution like the one below, which will not change the public store's API ?
> 
> ```
> RedisStore.prototype.incr = function(ip, cb) {
>   return this.incrementWithCustomTimeWindow(ip, this.timeWindow, cb);
> }
> RedisStore.prototype.incrementWithCustomTimeWindow = function (ip, timeWindow, cb) { ... }
> ```

@leorossi I do not see any problem for using `.child` method for creating a new store. If you have any problem, it will already pop up for `onRoute` hooks custom `routeOptions`. The method `.child` already take care of the logic for different options passed for different store.

For your solution, it do not solve that the `CustomStore` need to implement a new API to solve this issue.

**Comment by climba03003** [comment]
Let me explain more,

1. The function `makeParams` will merge any custom options with the global options
2. `store.child` will create a new store with the option make by `makeParams`
3. You need to clone `pluginComponent` and replace `.store` created on `step 2`.

With the upper step, you will have a new `pluginComponent` passed inside `rateLimitRequestHandler`. That means no logic inside `rateLimitRequestHandler` need to be changed and the store already applied the new option.

Does it clear to you @leorossi ?

**Comment by leorossi** [comment]
Yes, it's clear, problem is that in `rateLimitRequestHandler` I don't have access to `makeParams`. And if I change the store inside the `decorate` function, two different routes with different configuration in `preHandler` won't work because the last one will override the "global" store.

So basically I cannot call `child` inside `rateLimitRequestHandler`.

I can either do something like this
```js
if (params.timeWindow && pluginComponent.store.timeWindow !== params.timeWindow) {
          _store = Object.assign({}, pluginComponent.store)
          _store.incr = pluginComponent.store.incr
          _store.child = pluginComponent.store.child
          _store.timeWindow = params.timeWindow
        }
        _store.incr(...)
```

or maybe something like this
```js
const _oldTimeWindow = store.timeWindow;
if (params.timeWindow && pluginComponent.store.timeWindow !== params.timeWindow) {
  store.timeWindow = params.timeWindow
}

store.incr(...)
store.timeWindow = _oldTimeWindow
```

climba03003:
@leorossi Changes inside the `decorate` function will be affect the global instance as we need to call `Object.create` which means it will create a full clone of new object. Then, the store instance is encapsulated for that particular decorate inner function.

leorossi:
@climba03003 last commit is a new approach where the store is cloned if there is a custom `timeWindow` deifned.

If you're happy with that I will make a specific unit test for this scenario

climba03003:
> 
> 
> @climba03003 last commit is a new approach where the store is cloned if there is a custom `timeWindow` deifned.
> 
> If you're happy with that I will make a specific unit test for this scenario

My first glance is the example from document will not work on this case. Is there any strong stance on not using the existing `.child` method which must be provided for must `Store`?

```js
function CustomStore (options) {
  this.options = options
  this.current = 0
}

CustomStore.prototype.incr = function (key, cb) {
  const timeWindow = this.options.timeWindow
  this.current++
  cb(null, { current: this.current, ttl: timeWindow - (this.current * 1000) })
}

CustomStore.prototype.child = function (routeOptions) {
  // We create a merged copy of the current parent parameters with the specific
  // route parameters and pass them into the child store.
  const childParams = Object.assign(this.options, routeOptions)
  const store = new CustomStore(childParams)
  // Here is where you may want to do some custom calls on the store with the information
  // in routeOptions first...
  // store.setSubKey(routeOptions.method + routeOptions.url)
  return store
}
```

leorossi:
ok it makes sense, I updated the code and added a test for this use case

index.js
```diff
@@ -96,14 +96,17 @@ async function rateLimitPlugin (fastify, settings) {
...
-
   if (!fastify.hasDecorator('rateLimit')) {
     // The rate limit plugin can be registered multiple times but decorate throws if called multiple times for the same field
     fastify.decorate('rateLimit', function rateLimit (options) {
       let params = globalParams
       if (options) {
         params = makeParams(options)
       }
+      if (params.timeWindow && params.timeWindow !== pluginComponent.store.timeWindow) {
+        const child = pluginComponent.store.child(Object.assign({}, { routeInfo: {} }, params))
```
climba03003, line 107: Need to clone `pluginComponent` before replacing the property. Otherwise, it is modifying the global one.

Issue #191, opened 2021-12-06
More efficient limiting


vader-sama:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### Description

I ​think the current rate limiter is missing the point. Rate limiters are made to prevent brute force attacks, right?
The thing is, `fastify-rate-limiter` only blocks **most** of brute force requests.
Let me give you an example:
I've reached the limit and now I'm locked for 2 minutes. I'll wait for 1 minute and will send a request to the server again.
It will says that *I should wait for 1 more minute*.

Nothing wrong right? 🧐

But imagine a script; which will send 15 requests to the server every minute. It will reach the rate limit after 1 minute (lets assume limit is at `15` and lock time is `2m`). It will keep sending requests until the rate limit ends. after that it will send another 15 requests to the server. so its sending 15 requests every 3 minute even when rate limiter is on!

###  Solution
I think the best solution is to renew the lock time on every request when the ip is locked

### Motivation

I've discovered `Fastify` recently, and since then I'm in love with it 😁 

I try to use `Fastify` version  of everything when I'm developing using `Fastify`! Rate limiter is one of those. Maybe the feature that I'm requesting is not useful But I needed it, so I shared my thoughts with you guys.

Eomm:
Would you like to send a Pull Request to address this feature? Remember to add unit tests.

I think the `onExceeding` function could be improved to support this

vader-sama:
@Eomm Sure why not
I'll look into this at weekend
Thanks for the response! 🙏🤩

jsumners:
I wonder how others handle this. For example, if you exceed the limit on the GitHub API, do they extend you by another hour every time you accidentally hit the API when you're over the limit? I think there is some sort of windowing that must happen.

vader-sama:
@jsumners I believe time limits are mostly short and usually in minutes. Because locking someone for an hour doesn't make sense; But that's a good point tho. Making it dynamic is definitely much better, but How? I mean what's the scenario?

jsumners:
I don't have an answer except to say the logic for extending the ban should be modular such that it can be replaced by user desired logic.

vader-sama:
@jsumners We can get a boolean value from the user for it, and add more strategies in the future

vader-sama:
Was added in 67b4247a6da33df7d80146de7fa53707c1a07c2a

PR #193, opened 2021-12-09
feat(#191): Add another limiting strategy

vader-sama:
#### Checklist

- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)


climba03003 (changes requested):
It is actually forcing people to create more function in their custom store which should not be happen.

I think we can expose different official store for different strategy if they really need it.

test/global-rate-limit.test.js
```diff
@@ -948,3 +948,77 @@ test('exposeHeadRoutes', async t => {
...
+      method: 'GET'
+    })
+
+    t.equal(first.statusCode, 200)
+
+    t.equal(second.statusCode, 429)
+    t.equal(second.headers['x-ratelimit-limit'], 1)
+    t.equal(second.headers['x-ratelimit-remaining'], 0)
+    t.equal(second.headers['x-ratelimit-reset'], 5)
+  } catch (error) {
```
mcollina, line 981: remove this try/catch

test/global-rate-limit.test.js
```diff
@@ -948,3 +948,77 @@ test('exposeHeadRoutes', async t => {
...
```
mcollina, line 1018: remove the try/catch

test/route-rate-limit.test.js
```diff
@@ -1322,3 +1322,89 @@ test('Allow custom timeWindow in preHandler', t => {
...
+      method: 'GET'
+    })
+
+    t.equal(first.statusCode, 200)
+
+    t.equal(second.statusCode, 429)
+    t.equal(second.headers['x-ratelimit-limit'], 1)
+    t.equal(second.headers['x-ratelimit-remaining'], 0)
+    t.equal(second.headers['x-ratelimit-reset'], 5)
+  } catch (error) {
```
mcollina, line 1361: remove the try/catch

test/route-rate-limit.test.js
```diff
@@ -1322,3 +1322,89 @@ test('Allow custom timeWindow in preHandler', t => {
...
```
mcollina, line 1404: remove the try/catch

test/route-rate-limit.test.js
```diff
@@ -1322,3 +1322,89 @@ test('Allow custom timeWindow in preHandler', t => {
...
+    t.error(error)
+  }
+
+  redis.flushall(noop)
+  redis.quit(noop)
```
mcollina, line 1409: move these into a `t.teardown` function

mcollina (commented):
good work!

vader-sama:
@climba03003 That's correct but they are coding a lot already (creating `incr` function), I think requested feature isn't in scope of this PR and It needs another PR; Don't you agree? 😉

climba03003:
> @climba03003 That's correct but they are coding a lot already (creating `incr` function), I think requested feature isn't in scope of this PR and needs its own PR; Don't you agree? 😉

My suggest is preventing breaking change, it will break the people using `CustomStore` if they don't implement `incrAndRenew`. 


> @climba03003 That's correct but they are coding a lot already (creating `incr` function), I think requested feature isn't in scope of this PR and needs its own PR; Don't you agree? 😉

If we can provide a solution on `non-breaking-change`, I will go for that route. You can update the `storage` internal to provide the same feature.

I am against implementating one function for the store because it does not make sense on `CustomStore`.

vader-sama:
> > @climba03003 That's correct but they are coding a lot already (creating `incr` function), I think requested feature isn't in scope of this PR and needs its own PR; Don't you agree? 😉
> 
> If we can provide a solution on `non-breaking-change`, I will go for that route. You can update the `storage` internal to provide the same feature.
> 
> I am against implementating extra function for the store because it does not make sense on `CustomStore` use-case.

But what's the solution then?

climba03003:
> 

After https://github.com/fastify/fastify-rate-limit/blob/master/index.js#L78, you can pass the option inside the store.
`pluginComponent.store.continueExceeding = params.continueExceeding` 
Then, just update `.incr` function in the two store we provide.

climba03003:
Even better solution is that, update the `constructor` of the two `store` and accept new param.
Then, update the `.child` method that pass the `param`
Finally, update the `.incr` function.
If also allow us to `enable` this feature per route.

vader-sama:
@climba03003 What do you think now? (879ba4e57d2520)

vader-sama:
@climba03003 What do you think now? (17dbd486b)

climba03003 (approved):
LGTM. Thanks.

mcollina (approved):
lgtm

vader-sama:
@climba03003 Hi, this might be a weird question LOL; but is there any problem that my PR hasn't been merged yet?

Eomm (approved):

PR #204, opened 2022-02-08
chore: remove specific version from initiating redis for local tests

rluvaton:
#### Checklist

-----

I've done this for the following reasons:
1. The GitHub CI use the latest tag for Redis so I'm syncing the versions so no weird problem occurs when the local machine and the CI uses 2 different versions (I think it's best to pin to a specific version)
2. The current version doesn't have an ARM image which is a problem with the rise of the M1 CPUs (which is how I notice the problem)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "standard",
     "lint:fix": "standard --fix",
-    "redis": "docker run -p 6379:6379 --rm redis:3.0.7",
+    "redis": "docker run -p 6379:6379 --rm redis",
     "test": "standard && tap test/*.test.js && npm run typescript",
     "typescript": "tsd"
   },


package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "standard",
     "lint:fix": "standard --fix",
-    "redis": "docker run -p 6379:6379 --rm redis:3.0.7",
+    "redis": "docker run -p 6379:6379 --rm redis",
```
jsumners, line 9: This implies a significant jump in Redis versions. Current version is `redis:6`. I would think that this change incurs a semver major release. To avoid that, this should be changed to `redis:3`.

For the major, this should be changed to `redis:6` or `redis:6-alpine`.

rluvaton: I understand and agree with you but this is already what has been used in the GitHub CI:
https://github.com/fastify/fastify-rate-limit/blob/aef5e266bc6c07918bbfbaf73e910f377fd683d9/.github/workflows/ci.yml#L23

mcollina (approved):
lgtm

package.json
mcollina: I think this is ok.

Eomm (approved):

PR #208, opened 2022-02-09
Fixing `continueExceeding` option does not reset rate-limit - #207

### LINKED ISSUE #207: `continueExceeding` option does not reset rate-limit
### Prerequisites

### Fastify version

3.27.1

### Plugin version

5.7.0

### Node.js version

16.14.0

### Operating system

Windows

10

### Description

Restating what was said in https://github.com/fastify/fastify-rate-limit/issues/206:

My understanding is that when `continueExceeding` is set to `true`, the following scenario will occur:

- User hits rate limit and is limited for a certain amount of time (say 60 seconds)
- If the user makes another request between the time they were rate-limited and the time their rate-limit was due to expire, then it is reset back to 60 seconds

However, this does not appear to be the case when testing, and the rate-limit is not reset.

### Steps to Reproduce

1. Stand up Fastify server:

```js
const fastify = require("fastify")({ logger: true });

// Import plugins
const rateLimit = require("fastify-rate-limit");

/**
 * @description test function
 */
async function server() {
	fastify.register(rateLimit, {
		max: 2,
		timeWindow: 10000,
		continueExceeding: true,
	});

	fastify.route({
		method: "GET",
		url: "/foo",
		handler: async (req, res) => {
			res.send("ok");
		},
	});
	await fastify.listen({ port: 8000 });
}

server();
```

2. Make GET request twice to trigger rate-limit:

```
Request:

> GET /foo HTTP/1.1
> Host: localhost:8000
> Accept: */*

Response:

< HTTP/1.1 200 OK
< x-ratelimit-limit: 2
< x-ratelimit-remaining: 1
< x-ratelimit-reset: 10
< content-type: text/plain; charset=utf-8
< content-length: 2
< Date: Wed, 09 Feb 2022 15:10:55 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
```

```
Request:

> GET /foo HTTP/1.1
> Host: localhost:8000
> Accept: */*

Response:

< HTTP/1.1 200 OK
< x-ratelimit-limit: 2
< x-ratelimit-remaining: 0
< x-ratelimit-reset: 9
< content-type: text/plain; charset=utf-8
< content-length: 2
< Date: Wed, 09 Feb 2022 15:10:56 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
```

3. Make GET request after hitting 2 request rate-limit:
```
Request:
> GET /foo HTTP/1.1
> Host: localhost:8000
> Accept: */*

Response:
< HTTP/1.1 429 Too Many Requests
< x-ratelimit-limit: 2
< x-ratelimit-remaining: 0
< x-ratelimit-reset: 8
< retry-after: 10000
< content-type: application/json; charset=utf-8
< content-length: 99
< Date: Wed, 09 Feb 2022 15:10:57 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
```
4. Make GET request within 10 seconds after rate-limit triggered (still expecting to be denied access, as step 3. should have reset the rate-limit back to 10 seconds, so shouldn't have allowed access again until 15:11:07 GMT):

```
Request:
> GET /foo HTTP/1.1
> Host: localhost:8000
> Accept: */*

Response:
< HTTP/1.1 200 OK
< x-ratelimit-limit: 2
< x-ratelimit-remaining: 1
< x-ratelimit-reset: 10
< content-type: text/plain; charset=utf-8
< content-length: 2
< Date: Wed, 09 Feb 2022 15:11:01 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
```

### Expected Behavior

Rate-limit should be reset if user makes request between the time they were rate-limited and the time their rate-limit was due to expire if `continueExceeding` set to `true`.

rluvaton:
also changed the package.json so tap would run all the tests in the test folder and it's sub directories

#### Checklist

fix #207

### DIFF
--- a/index.js
+++ b/index.js
@@ -186,6 +186,14 @@ function rateLimitRequestHandler (params, pluginComponent) {
     let current = 0
     let ttl = 0
 
+    let maximum
+
+    if (typeof params.max === 'number' && !isNaN(params.max)) {
+      maximum = params.max
+    } else {
+      maximum = await params.max(req, key)
+    }
+
     // As the key is not allowList in redis/lru, then we increment the rate-limit of the current request
     try {
       const res = await new Promise(function (resolve, reject) {
@@ -195,7 +203,7 @@ function rateLimitRequestHandler (params, pluginComponent) {
             return
           }
           resolve(res)
-        })
+        }, maximum)
       })
 
       current = res.current
@@ -204,14 +212,6 @@ function rateLimitRequestHandler (params, pluginComponent) {
       if (!params.skipOnError) {
         throw err
       }
-    }
-
-    let maximum
-
-    if (typeof params.max === 'number' && !isNaN(params.max)) {
-      maximum = params.max
-    } else {
-      maximum = await params.max(req, key)
     }
 
     const timeLeft = Math.floor(ttl / 1000)
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "tsd": "^0.14.0"
   },
   "dependencies": {
-    "@sinonjs/fake-timers": "^9.1.0",
     "fastify-plugin": "^3.0.0",
     "ms": "^2.1.1",
     "tiny-lru": "^8.0.1"
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "tsd": "^0.14.0"
   },
   "dependencies": {
+    "@sinonjs/fake-timers": "^9.1.0",
     "fastify-plugin": "^3.0.0",
     "ms": "^2.1.1",
     "tiny-lru": "^8.0.1"
--- a/store/LocalStore.js
+++ b/store/LocalStore.js
@@ -9,17 +9,22 @@ function LocalStore (timeWindow, cache, app, continueExceeding) {
   this.continueExceeding = continueExceeding
 }
 
-LocalStore.prototype.incr = function (ip, cb) {
+LocalStore.prototype.incr = function (ip, cb, max) {
   const nowInMs = Date.now()
   const current = this.lru.get(ip) || { count: 0, iterationStartMs: nowInMs }
 
   current.count++
 
-  this.lru.set(ip, current)
-
   if (this.continueExceeding) {
+    if (current.count > max) {
+      this.lru.delete(ip)
+    }
+
+    // It will recalculate the TTL if the item is missing - count exceeded the maximum
+    this.lru.set(ip, current)
     cb(null, { current: current.count, ttl: this.timeWindow })
   } else {
+    this.lru.set(ip, current)
     cb(null, { current: current.count, ttl: this.timeWindow - (nowInMs - current.iterationStartMs) })
   }
 }
--- a/test/github-issues/issue-207.test.js
+++ b/test/github-issues/issue-207.test.js
@@ -1,13 +1,20 @@
 'use strict'
 
+const FakeTimers = require('@sinonjs/fake-timers')
 const t = require('tap')
 const test = t.test
 const Fastify = require('fastify')
 const rateLimit = require('../../index')
 
-test('issue #207 - when continueExceeding is true and the store is local then it should reset the rate-limit', async t => {
-  const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
+t.beforeEach(t => {
+  t.context.clock = FakeTimers.install()
+})
 
+t.afterEach(t => {
+  t.context.clock.uninstall()
+})
+
+test('issue #207 - when continueExceeding is true and the store is local then it should reset the rate-limit', async t => {
   const fastify = Fastify()
 
   fastify.register(rateLimit, {
@@ -35,7 +42,7 @@ test('issue #207 - when continueExceeding is true and the store is local then it
     method: 'GET'
   })
 
-  await sleep(3000)
+  t.context.clock.tick(3000)
 
   const secondRateLimitWithResettingTheRateLimitTimer = await fastify.inject({
     url: '/',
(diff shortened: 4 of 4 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "lint": "standard",
     "lint:fix": "standard --fix",
     "redis": "docker run -p 6379:6379 --rm redis",
-    "test": "standard && tap test/*.test.js && npm run typescript",
+    "test": "standard && tap && npm run typescript",
```
rluvaton, line 10: By default tap search for tests in the test folder that have the extension `.test.js` (or other criteria), I've removed the explicit files location as it wouldn't work for the location of the new test - `/test/github-issues/<file>`, I've could add another glob (`test/**/*.test.js`) but instead, I removed it

test/github-issues/issue-207.test.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const Fastify = require('fastify')
+const rateLimit = require('../../index')
+
+test('issue #207 - when continueExceeding is true and the store is local then it should reset the rate-limit', async t => {
+  const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
```
rluvaton, line 9: This is not very good as it will make the test longer, have any idea of how to solve this? is there some utility in `tap` for _updating_ time?

rluvaton:
So I've found the bug:

In the `LocalStore` there is a `beat` function that reset the counters for each IP every `timeWindow`, But! because there is only 1 interval for calling `beat` for everything then the counters reset for every IP.

What the user expects is that we reset the timer but this is a problem since we have **1** interval.

A possible solution would be:
1. When a client reaches the rate limit then we create a timer with the new time window that when the timer ends the `lru` will reset it request counter
2. when the client sends another request while still being rate limited we reset the timer to start from the start of the time window
3. when the client timer ends and the general beat interval tick (timer tick not Node tick 😉) and the user did not send any request in the middle the user is joined to the general beat interval 

this is not a very good solution as it does not scale properly and have a security vulnerability which the user can reach rate limit with multiple IPs or have a different value in the `x-forwarded-for` header

another solution is when a request is made after rate limit then we assign the client to the interval with the current jump (every 10ms an interval for beat and then when the user reach rate limit we add him to the right interval, with this solution the number of intervals does not depend on the number of users

I'm sorry if my English is not the best 😬


----

When writing the possible solution I think I found another bug or is this the expected behavior?:

Let's say the time window is 60s and the rate limit is 1
User 1: send a request (which starts the global interval) - ok
User 2: send a request after 50 seconds - ok
User 2: send another request after a second - rate limited
User 2: send another request after 15 seconds - not been rate limited 

-----

@mcollina @Fdawgs What do you think?

test/github-issues/issue-207.test.js
Eomm: We used https://sinonjs.org/releases/latest/fake-timers/ in others repos

Fdawgs:
> another solution is when a request is made after rate limit then we assign the client to the interval with the current jump (every 10ms an interval for beat and then when the user reach rate limit we add him to the right interval, with this solution the number of intervals does not depend on the number of users

Sounds good!

rluvaton:
@Fdawgs How can we have the max without passing it to the store (or we can pass it as it's not a breaking change)?

index.js
```diff
@@ -186,6 +186,14 @@ function rateLimitRequestHandler (params, pluginComponent) {
...
     let ttl = 0
 
+    let maximum
+
+    if (typeof params.max === 'number' && !isNaN(params.max)) {
+      maximum = params.max
+    } else {
+      maximum = await params.max(req, key)
+    }
+
```
rluvaton, line 196: Would this break if moving the check before rather than after? it shouldn't... we don't make any promises when the function would be executed...

store/LocalStore.js
```diff
@@ -9,17 +9,22 @@ function LocalStore (timeWindow, cache, app, continueExceeding) {
   this.continueExceeding = continueExceeding
 }
 
-LocalStore.prototype.incr = function (ip, cb) {
+LocalStore.prototype.incr = function (ip, cb, max) {
```
rluvaton, line 12: adding the `max` here although it's not ideal because:
1. I can't pass it in the constructor as `max` can be a function that needs the request which we don't have/want access
2. passing it here and not before the `cb` as it would breaking change and we're tying to avoid it

mcollina (approved):
lgtm

store/LocalStore.js
rluvaton: What that function would do?

Eomm: 1. you don't need to move the `maximum` logic in index.js
2. you don't need to provide the `max` arg to the localstore file

rluvaton: What would that function do I'm trying to understand, is it gonna check if `continueExceeding` is true and if so it would recalculate the TTL?

I'm thinking of a name for that function, maybe `recalculateTTLIfNeeded`? Or `recalculateTTL` and it wouldn't check for `continueExceeding` and we would check that in the place where we call that function

WDYT?

Fdawgs (approved):

store/LocalStore.js
Eomm: The `index.js` file has the `current.coun` and the `max` values.

It would be something like

```
    if (current.count > max && theStore.forceDelete) {
      theStore.forceDelete(ip)
    }
```

Eomm (changes requested):

store/LocalStore.js
rluvaton: > The `index.js` file has the `current.coun` and the `max` values.
> 
> 
> 
> It would be something like
> 
> 
> 
> ```
> 
>     if (current.count > max && theStore.forceDelete) {
> 
>       theStore.forceDelete(ip)
> 
>     }
> 
> ```

@Eomm sorry for the late reply

This would cause a problem as the next increment would start from 0

Fdawgs:
I think this one is good to go

PR #209, opened 2022-02-09
fix: add NaN handling

rluvaton:
#### Checklist

When checked for number using `typeof <var> === 'number'` then forgot to handle `NaN` which is type `number` but invalid value, currently just handled `NaN` as invalid value but not throwed/warn anything, which is worse IMO so I can change it if needed

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,14 +48,14 @@ async function rateLimitPlugin (fastify, settings) {
   globalParams.labels = labels
 
   // define the global maximum of request allowed
-  globalParams.max = (typeof settings.max === 'number' || typeof settings.max === 'function')
+  globalParams.max = ((typeof settings.max === 'number' && !isNaN(settings.max)) || typeof settings.max === 'function')
     ? settings.max
     : 1000
 
   // define the global Time Window
   globalParams.timeWindow = typeof settings.timeWindow === 'string'
     ? ms(settings.timeWindow)
-    : typeof settings.timeWindow === 'number'
+    : typeof settings.timeWindow === 'number' && !isNaN(settings.timeWindow)
       ? settings.timeWindow
       : 1000 * 60
 
@@ -208,7 +208,7 @@ function rateLimitRequestHandler (params, pluginComponent) {
 
     let maximum
 
-    if (typeof params.max === 'number') {
+    if (typeof params.max === 'number' && !isNaN(params.max)) {
       maximum = params.max
     } else {
       maximum = await params.max(req, key)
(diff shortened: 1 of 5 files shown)


Eomm (changes requested):
Could you add a test?

rluvaton:
Shouldn't we throw an error and then it will make more sense to test it, no?

Eomm:
I would keep the old behaviour for now

rluvaton:
I've added the tests for global `max` and global `timeWindow`

test/global-rate-limit.test.js
```diff
@@ -98,6 +99,48 @@ test('With text timeWindow', async t => {
...
+
+  t.equal(res.statusCode, 200)
+  t.equal(res.headers['x-ratelimit-reset'], defaultTimeWindowInSeconds)
+
+  res = await fastify.inject('/')
+
+  t.equal(res.statusCode, 429)
+
+  // Wait for almost 60s to make sure the time limit is right
+  t.context.clock.tick(55_000)
```
darkgl0w, line 126:
```suggestion
  t.context.clock.tick(55000)
```
that's not supported under Node.js 10

test/global-rate-limit.test.js
```diff
@@ -98,6 +99,48 @@ test('With text timeWindow', async t => {
...
+
+  res = await fastify.inject('/')
+
+  t.equal(res.statusCode, 429)
+
+  // Wait for seconds left until the time limit reset
+  t.context.clock.tick(5_000)
```
darkgl0w, line 133:
```suggestion
  t.context.clock.tick(5000)
```

index.js
```diff
@@ -208,7 +208,7 @@ function rateLimitRequestHandler (params, pluginComponent) {
 
     let maximum
 
-    if (typeof params.max === 'number') {
+    if (typeof params.max === 'number' && !isNaN(params.max)) {
```
rluvaton, line 211: Do you want me to add a test that the library fail or change the code here, I did not add a test for the `max` property for specific route as it would fail the library as it tries to call `NaN` as a function

darkgl0w (commented):
Nits.

test/global-rate-limit.test.js
rluvaton: Thank you, I saw that in the CI my new commit don't use them

test/global-rate-limit.test.js
rluvaton: Thank you, I saw that in the CI my new commit don't use them

rluvaton:
I'll fix the failed test in a new PR  (the failed test was `With enabled IETF Draft Spec` and has nothing to do with this change

rluvaton:
> could you rebase?

Done!

mcollina (approved):
lgtm

Eomm (approved):

PR #217, opened 2022-02-12
fix(local-store): local store should have time window for each user

### LINKED ISSUE #215: LocalStore behavior is different that RedisStore
### Prerequisites

### Fastify version

3.27.1

### Plugin version

5.7.0

### Node.js version

16.14.0

### Operating system

macOS

12.2

### Description
In the `LocalStore` the `timeWindow` **time**  (not the amount of requests a user can make) is shared for all the users and in the `RedisStore` each user has its own `timeWindow`

_Originally posted by @rluvaton in https://github.com/fastify/fastify-rate-limit/issues/208#issuecomment-1034169203_

### Steps to Reproduce

**when using the local store:**

Let's say the time window is 10s and the rate limit is 1 and `continueExceeding` is false
1. User 1: send a request (which internally starts the global interval) - ok
2. User 2: send a request after 7 seconds - ok
3. User 2: send another request again - rate limited
4. User 2: send another request after 5 seconds - not been rate-limited (which he should as 10 has not passed for him only 5)

From reading the Redis store code and testing it, each user has its own time window (meaning User 2 in step 4 would be rate-limited)

[Reproduction repo](https://github.com/rluvaton/fastify_fastify-rate-limit-repro-215) just follow the steps to run or scroll down to see the final result for me

### Expected Behavior

Both stores to have the same behavior

rluvaton:
#### Checklist

Fix #215

This PR effect #207 but not fixes it, see https://github.com/fastify/fastify-rate-limit/pull/217/files#r805223197


store/LocalStore.js
```diff
@@ -3,37 +3,27 @@
...
-  this.lru.set(ip, ++current)
+  const current = this.lru.get(ip) || { count: 0, iterationStartMs: Date.now() }
+  const isNewItem = current.count === 0
+
+  current.count++
+
+  // We want to update the TTL only if it's a new item (because it doesn't have any TTL) or `continueExceeding` is true
+  // Passing false to the 3rd parameter = update the TTL AND I don't know if the item with that key already exist
+  // Passing true means don't update the TTL AND I'm sure there is an item with that key
+  this.lru.set(ip, current, !(isNewItem || this.continueExceeding))
```
rluvaton, line 21: I think we only want to update the TTL when the user has already reached his rate limit, right? otherwise, it would never be deleted...

how should we approach this? as the store doesn't know when the client reached its maximum request in the time window...

I can update the TTL only when an item is a new item and deal with this error later in the #208

store/LocalStore.js
```diff
@@ -3,37 +3,27 @@
...
-  this.lru.set(ip, ++current)
+  const current = this.lru.get(ip) || { count: 0, iterationStartMs: Date.now() }
+  const isNewItem = current.count === 0
+
+  current.count++
+
+  // We want to update the TTL only if it's a new item (because it doesn't have any TTL)
+  // Passing false to the 3rd parameter = update the TTL AND I don't know if the item with that key already exist
+  // Passing true means don't update the TTL AND I'm sure there is an item with that key
+  this.lru.set(ip, current, !isNewItem)
```
rluvaton, line 21: We should also update the TTL when `continueExceeding` is `true` AND the user has already reached his rate limit, right?

I'm keeping the bug as it's the same bug as #207 because:
1. #215 bug is more important IMO
2. this is not the PR for that, even though I'm changing the related code we should think about how to approach this as the store doesn't know when the client reached its maximum request in the time window...

store/LocalStore.js
```diff
@@ -3,37 +3,28 @@
...
-  function beat () {
-    this.lru.clear()
-    this.msLastBeat = null
-  }
 }
 
 LocalStore.prototype.incr = function (ip, cb) {
-  let current = this.lru.get(ip) || 0
-  this.lru.set(ip, ++current)
+  const nowInMs = Date.now()
```
rluvaton, line 13: Done this because some tests failed as `x-ratelimit-reset` was not exactly the same second (as it is later rounded):
https://github.com/fastify/fastify-rate-limit/blob/4ad50c6915176db0d7a9ce4cae3f5a2f5a87d82b/test/route-rate-limit.test.js#L46

store/LocalStore.js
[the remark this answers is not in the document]
Eomm: Why it is not documented?
I would not trust about a hidden feature that may be removed or not tested at all

Eomm (commented):
LGTM, but I would check if the tiny-lru's secret `bypass` parameter can be used safely

store/LocalStore.js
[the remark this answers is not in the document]
rluvaton: I totally agree with you, but this was the best solution, I even looked at other LRU libraries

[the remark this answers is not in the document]
Eomm: The best thing is to ask the maintainer: maybe it is just a documentation's bug.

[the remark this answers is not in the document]
rluvaton: That's what I did - https://github.com/avoidwork/tiny-lru/issues/47
Still waiting for the response...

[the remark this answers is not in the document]
rluvaton: So he responded and was kind to fix the problem that the item TTL is being recomputed and release a new breaking version - `v8.0.0`, waiting for the NPM package to release and then gonna fix this

rluvaton:
I fix your comment, thanks!

test/github-issues/issue-215.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+t.beforeEach(t => {
+  t.context.clock = FakeTimers.install()
+})
+
+t.afterEach(t => {
+  t.context.clock.uninstall()
+})
+
+test('issue #215 - when using local store, 2nd user is not being rate limited when time window is passed for the 1st user', async t => {
```
Fdawgs, line 17: Not 100% on the test description, a tad long?

Fdawgs (approved):

test/github-issues/issue-215.test.js
rluvaton: Actually it should be:
```suggestion
test('issue #215 - when using local store, 2nd user should not be rate limited when time window is passed for the 1st user', async t => {
```

but it's still long, do you have a suggestion?

Fdawgs: That's clearer!

rluvaton:
Could you please re-run the tests 😀

package.json
```diff
@@ -7,7 +7,7 @@
     "lint": "standard",
     "lint:fix": "standard --fix",
     "redis": "docker run -p 6379:6379 --rm redis",
-    "test": "standard && tap test/*.test.js && npm run typescript",
+    "test": "standard && tap test/*.test.js test/**/*.test.js && npm run typescript",
```
darkgl0w, line 10:
```suggestion
    "test": "standard && tap \"test/**/*.test.js\" && npm run typescript",
```

darkgl0w (commented):
LGTM

package.json
darkgl0w: Yes it works as intended when escaping the quotes. Tested with Archlinux, Windows 10 & 11 (I can test with MacOS when I get back home).

What's your working environment ?

NB: the `not-found-handler-rate-limited.js` test file is currently not run. It needs to be renamed to something like `not-found-handler-rate-limited.test.js`

rluvaton: > Yes it works as intended when escaping the quotes. Tested with Archlinux, Windows 10 & 11 (I can test with MacOS when I get back home).
> 
> 
> 
> What's your working environment ?

Mac OS 😅

> NB: the `not-found-handler-rate-limited.js` test file is currently not run. It needs to be renamed to something like `not-found-handler-rate-limited.test.js`

Weird I didn't touch that file

darkgl0w: https://github.com/darkgl0w/fastify-rate-limit/runs/5188076153?check_suite_focus=true
Triggered a CI run on Mac OS, it works as expected with the escaped quotes. :P

rluvaton: Great! Thanks!

Eomm (approved):

PR #261, opened 2022-09-14
test: fix flaky redis unit tests

Uzlopak:
We were running the unit tests in a very flaky way. Instead of actually waiting the redis flushall command we were having a noop callback. So it could happen, that flushall was not done and we were already running the next test.

This PR fixes that by implementing a teardown for those redis tests and fixing the corresponding assertions. 

#### Checklist

### DIFF
--- a/test/global-rate-limit.test.js
+++ b/test/global-rate-limit.test.js
@@ -6,7 +6,6 @@ const Redis = require('ioredis')
 const Fastify = require('fastify')
 const rateLimit = require('../index')
 const FakeTimers = require('@sinonjs/fake-timers')
-const noop = () => { }
 const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms))
 
 const REDIS_HOST = '127.0.0.1'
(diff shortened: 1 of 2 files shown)


test/route-rate-limit.test.js
```diff
@@ -335,18 +337,20 @@ test('Skip on redis error', async t => {
   t.equal(res.headers['x-ratelimit-limit'], 2)
   t.equal(res.headers['x-ratelimit-remaining'], 1)
 
-  redis.flushall(noop)
-  await redis.quit()
```
Eomm, line 339: the scope of this test is to verify the `skipOnError` flag, so without quitting the redis connection we are not testing it

test/global-rate-limit.test.js
```diff
@@ -358,13 +357,16 @@ test('With redis store', async t => {
...
   t.equal(res.headers['x-ratelimit-remaining'], 1)
   t.equal(res.headers['x-ratelimit-reset'], 1)
+
+  t.teardown(async () => {
+    await redis.flushall()
+    await redis.quit()
+  })
 })
 
 test('Skip on redis error', async t => {
```
climba03003, line 372: This test is doing what it suppose to do. We need another way to simulate Redis error.

climba03003 (changes requested):

test/global-rate-limit.test.js
Uzlopak: I patched it now according to @Eomm remarks. 

Is it not acceptable?

climba03003: I believe we are just commenting on the same thing.
The patch appear later than my comment.

Eomm (approved):

PR #266, opened 2022-09-14
96% test coverage

Uzlopak:
Well 97% because I have no clue rn to test the error cases of RedisStore. 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -159,7 +159,7 @@ async function rateLimitPlugin (fastify, settings) {
 }
 
 async function buildRouteRate (pluginComponent, params, routeOptions) {
-  const hook = params.hook || defaultHook
+  const hook = params.hook
   const hookHandler = rateLimitRequestHandler(params, pluginComponent)
   if (Array.isArray(routeOptions[hook])) {
     routeOptions[hook].push(hookHandler)
--- a/index.js
+++ b/index.js
@@ -159,7 +159,7 @@ async function rateLimitPlugin (fastify, settings) {
 }
 
 async function buildRouteRate (pluginComponent, params, routeOptions) {
-  const hook = params.hook
+  const hook = params.hook || defaultHook
   const hookHandler = rateLimitRequestHandler(params, pluginComponent)
   if (Array.isArray(routeOptions[hook])) {
     routeOptions[hook].push(hookHandler)
--- a/test/global-rate-limit.test.js
+++ b/test/global-rate-limit.test.js
@@ -149,7 +149,7 @@ test('When passing NaN to the timeWindow property then the timeWindow should be 
   })
 })
 
-test('With ips allowList, allowed ips should not result in rate limitting', async t => {
+test('With ips allowList, allowed ips should not result in rate limiting', async t => {
   t.plan(3)
   const fastify = Fastify()
   await fastify.register(rateLimit, {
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -159,7 +159,7 @@ async function rateLimitPlugin (fastify, settings) {
 }
 
 async function buildRouteRate (pluginComponent, params, routeOptions) {
-  const hook = params.hook || defaultHook
+  const hook = params.hook
```
Uzlopak, line 162: at this point hook is always set

test/global-rate-limit.test.js
```diff
@@ -149,7 +149,7 @@ test('When passing NaN to the timeWindow property then the timeWindow should be
   })
 })
 
-test('With ips allowList', async t => {
+test('With ips allowList, allowed ips should not result in rate limitting', async t => {
```
Fdawgs, line 152:
```suggestion
test('With ips allowList, allowed ips should not result in rate limiting', async t => {
```

test/global-rate-limit.test.js
```diff
@@ -172,6 +172,29 @@ test('With ips allowList', async t => {
   t.equal(res.statusCode, 200)
 })
 
+test('With ips allowList, not allowed ips should result in rate limitting', async t => {
```
Fdawgs, line 175:
```suggestion
test('With ips allowList, not allowed ips should result in rate limiting', async t => {
```

index.js
climba03003: Imagine something like this happen `Object.assign({}, { foo: '' }, { foo: undefined })`, it will be `undefined`.
Or somehow, some people use `undefined` to pretend as revert as default in `route` level configuration.

Uzlopak: Its not possible because buildRouteRate gets called mergedRateLimitParams

https://github.com/fastify/fastify-rate-limit/blob/cde089460c974a30d19696bb12bec9587dea2c65/index.js#L138

mergedRateLimitParams gets generatd with makeParams 

https://github.com/fastify/fastify-rate-limit/blob/cde089460c974a30d19696bb12bec9587dea2c65/index.js#L134

makeParams merges with globalParams

https://github.com/fastify/fastify-rate-limit/blob/cde089460c974a30d19696bb12bec9587dea2c65/index.js#L153

And globalParams gets already a fallback to defaultHook
https://github.com/fastify/fastify-rate-limit/blob/cde089460c974a30d19696bb12bec9587dea2c65/index.js#L64

So if you set hook to undefined on route level than it gets overwritten by the global.hook setting in makeParams.

There is no way that buildRouteRate would receive an option-object, where params.hook is undefined.

climba03003: It will be `undefined` because `Object.assign({}, { hook: defaultHook }, { hook: undefined })` is `{ hook: undefined }` 
`Object.assign` respect the last one only which is `routeParams`.

Uzlopak: I revert this change, I am ok with the increased coverage

climba03003: PoC test if you need. It is the case that what I am talking about. And it is how `Object.assign` works.

```js
t.plan(2)

const fastify = Fastify()
await fastify.register(rateLimit, {    
  max: 2,
  timeWindow: 1000,
  hook: 'preHandler'
})

fastify.addHook('onRoute', function(routeOptions) {
  t.equal(routeOptions.preHandler, undefined)
  t.equal(routeOptions.onRequest.length, 1)
})

fastify.get('/', {
  config: {
    rateLimit: {
      hook: undefined
    }
  },
  exposeHeadRoute: false
}, async (req, reply) => 'fastify is awesome !')

await fastify.ready()
```

Uzlopak: Is there a way to check if a hook is set or not?

Uzlopak: done

Eomm (approved):

PR #285, opened 2023-02-26
Don't set the reply code automatically

### LINKED ISSUE #284: Don't set the reply code automatically
### Prerequisites

- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

A proposal to modify the behavior of rate limit errors.

Since v4, `fastify-sensible` no longer has the error handler that sets the reply code automatically; `fastify-multipart` doesn't set it either when the Content Too Large error is thrown for instance — they only set the `error.statusCode`, throw, and let the dev decide what to do with that information. 

Currently, `rate-limit` sets the reply code as well as the error code and it breaks things like `reply.redirect("/")` calls in the error handler that don't specify a redirect code.

It's not a huge issue, but a minor inconsistency between plugins IMO.

What do you think?

### Motivation

_No response_

### Example

_No response_

gurgunday:
Closes #284

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ You can change the response by providing a callback to `errorResponseBuilder` or
 
 ```js
 fastify.setErrorHandler(function (error, request, reply) {
-  if (reply.statusCode === 429) {
+  if (error.statusCode === 429) {
+    reply.code(429)
     error.message = 'You hit the rate limit! Slow down please!'
   }
   reply.send(error)
--- a/index.js
+++ b/index.js
@@ -263,11 +263,11 @@ function rateLimitRequestHandler (params, pluginComponent) {
       params.onBanReach(req, key)
     }
 
-    throw params.errorResponseBuilder(respCtx)
+    throw params.errorResponseBuilder(req, respCtx)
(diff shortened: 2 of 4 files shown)


gurgunday:
I don't get the errors because they were failing before I even made any changes.

As far as I can tell, the changes I made caused no new failures. I also added a new one that would fail before this change.

gurgunday:
Ok, I had done something not smart, reverted that now. Should be good to go.

mcollina (approved):
lgtm

this is a semver-majorn change

bcomnes:
I have sensible and rate limit both set. Do I need to change anything to update to to v8?

gurgunday:
> I have sensible and rate limit both set. Do I need to change anything to update to to v8?

I have a similar setup and upgrading to v8 required no changes. If you have routes with a custom error handler where you don't set a reply code or use something like `reply.redirect()`, you might need to set the reply code yourself. For other routes, the error response should be identical to v7.

bcomnes:
I see, thank you!

PR #308, opened 2023-07-12
throttle

Uzlopak:
#### Checklist


mcollina (changes requested):
You should use pipeline instead, .pipe does not add error handling.

Note that using an intermediate stream adds significant overhead.

Uzlopak:
> You should use pipeline instead, .pipe does not add error handling.

Ok, I will use pipeline

> Note that using an intermediate stream adds significant overhead.

Should I put that into the readme? Or is there a different approach, that I am not aware of?

I actually wanted to write about the feature for the blog of @Eomm 

https://github.com/Uzlopak/blog/blob/main/case.md

Uzlopak:
@mcollina 

Do you think it makes sense to create a new package called @fastify/throttle?

index.js
```diff
@@ -158,6 +170,41 @@ async function fastifyRateLimit (fastify, settings) {
...
+  const bps = throttleOpts.bps
+  return function (request, reply, payload, done) {
+    if (payload && payload.pipe) {
+      const throttleStream = new ThrottleStream({ bps })
+      payload.pipe(throttleStream)
+      done(null, throttleStream)
+      return
+    } else if (Buffer.isBuffer(payload)) {
+      const throttleStream = new ThrottleStream({ bps })
+      Readable.from(payload).pipe(throttleStream)
```
mcollina, line 195: It does not have much sense to throttle in this case. Throlling is needed to limit the resource consumed by the caller. In this specific case, we are holding quite a bit of data in memory, literally increasing the resource consumed to process the request.

mcollina (commented):
I think a new module would be better.

---

Something that this does not handle is throttling across multiple request/multiple server for a given user.

Uzlopak:
I am actually working on the distributed part. 

see
https://github.com/Uzlopak/blog/blob/main/case.md#throttlegroups

also yes, i initially wanted to just handle streams. Maybe make it optional to handle also buffer and strings if an option is set (opt-in)

Uzlopak:
Created a new repo fastify/throttle and will do the changes there.

PR #313, opened 2023-07-31
chore: remove defaultOnBanReach helper fn

gurgunday:
All custom function parameters, like `onExceeded` and `onExceeding`, only get run if they're defined. However, `onBan` has a default function that is just empty (old, line 281, 268)

Under stress from an IP, that function can get called thousands of times, which probably still wouldn't cause a problem (🤣), but why run it in the first place?

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,8 @@ async function fastifyRateLimit (fastify, settings) {
   globalParams.allowList = settings.allowList || settings.whitelist || null
   globalParams.ban = settings.ban || null
   globalParams.onBanReach = typeof settings.onBanReach === 'function' ? settings.onBanReach : undefined
+  globalParams.onExceeding = typeof settings.onExceeding === 'function' ? settings.onExceeding : undefined
+  globalParams.onExceeded = typeof settings.onExceeded === 'function' ? settings.onExceeded : undefined
   globalParams.continueExceeding = typeof settings.continueExceeding === 'boolean' ? settings.continueExceeding : false
 
   // define the name of the app component. Related to redis, it will be use as a part of the keyname define in redis.
@@ -90,9 +92,6 @@ async function fastifyRateLimit (fastify, settings) {
   globalParams.keyGenerator = typeof settings.keyGenerator === 'function'
     ? settings.keyGenerator
     : (req) => req.ip
-
-  globalParams.onExceeded = typeof settings.onExceeded === 'function' ? settings.onExceeded : undefined
-  globalParams.onExceeding = typeof settings.onExceeding === 'function' ? settings.onExceeding : undefined
 
   // define if error message was overwritten with a custom error response callback
   if (typeof settings.errorResponseBuilder === 'function') {
--- a/index.js
+++ b/index.js
@@ -69,7 +69,7 @@ async function fastifyRateLimit (fastify, settings) {
   globalParams.allowList = settings.allowList || settings.whitelist || null
   globalParams.ban = settings.ban || null
   globalParams.onBanReach = typeof settings.onBanReach === 'function' ? settings.onBanReach : undefined
-  globalParams.continueExceeding = settings.continueExceeding || false
+  globalParams.continueExceeding = typeof settings.continueExceeding === 'boolean' ? settings.continueExceeding : false
 
   // define the name of the app component. Related to redis, it will be use as a part of the keyname define in redis.
   const pluginComponent = {
@@ -91,8 +91,8 @@ async function fastifyRateLimit (fastify, settings) {
     ? settings.keyGenerator
     : (req) => req.ip
 
-  globalParams.onExceeded = settings.onExceeded
-  globalParams.onExceeding = settings.onExceeding
+  globalParams.onExceeded = typeof settings.onExceeded === 'function' ? settings.onExceeded : undefined
+  globalParams.onExceeding = typeof settings.onExceeding === 'function' ? settings.onExceeding : undefined
 
   // define if error message was overwritten with a custom error response callback
   if (typeof settings.errorResponseBuilder === 'function') {
@@ -236,14 +236,12 @@ function rateLimitRequestHandler (params, pluginComponent) {
       if (params.addHeadersOnExceeding[params.labels.rateRemaining]) { res.header(params.labels.rateRemaining, maximum - current) }
       if (params.addHeadersOnExceeding[params.labels.rateReset]) { res.header(params.labels.rateReset, timeLeft) }
 
-      if (typeof params.onExceeding === 'function') {
-        params.onExceeding(req, key)
-      }
+      params.onExceeding?.(req, key)
+
       return
     }
-    if (typeof params.onExceeded === 'function') {
-      params.onExceeded(req, key)
-    }
+
+    params.onExceeded?.(req, key)
 
     if (params.addHeaders[params.labels.rateLimit]) { res.header(params.labels.rateLimit, maximum) }
     if (params.addHeaders[params.labels.rateRemaining]) { res.header(params.labels.rateRemaining, 0) }
@@ -263,7 +261,7 @@ function rateLimitRequestHandler (params, pluginComponent) {
 
     if (code === 403) {
       respCtx.ban = true
-      if (typeof params.onBanReach === 'function') params.onBanReach(req, key)
+      params.onBanReach?.(req, key)
     }
 
     throw params.errorResponseBuilder(req, respCtx)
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -265,7 +263,7 @@ function rateLimitRequestHandler (params, pluginComponent) {
 
     if (code === 403) {
       respCtx.ban = true
-      params.onBanReach(req, key)
+      if (typeof params.onBanReach === 'function') params.onBanReach(req, key)
```
Uzlopak, line 266: Uff. I dont think this is nice. the noop can be optimized by v8, but this is basically a string compare, and is probably slower.

gurgunday: I see, I though about function inlining even though that parameter is dynamic, which made me unsure

What about these then? Wouldn't replacing them with empty functions too better then?
```js
      if (typeof params.onExceeding === 'function') {
        params.onExceeding(req, key)
      }
      return
    }
    if (typeof params.onExceeded === 'function') {
      params.onExceeded(req, key)
    }
```

index.js
```diff
@@ -94,23 +91,24 @@ async function fastifyRateLimit (fastify, settings) {
...
   if (typeof settings.errorResponseBuilder === 'function') {
     globalParams.errorResponseBuilder = settings.errorResponseBuilder
     globalParams.isCustomErrorMessage = true
+  } else {
+    globalParams.errorResponseBuilder = defaultErrorResponse
+    globalParams.isCustomErrorMessage = false
   }
 
-  globalParams.skipOnError = settings.skipOnError || false
+  globalParams.skipOnError = typeof settings.skipOnError === 'boolean' ? settings.skipOnError : false
```
Uzlopak, line 106: if you want to default to false, it is perfectly fine to do `|| false` If you want to default to true, than I would say, this makes sense, but still we would have the "new" approach to do `?? true`.
Or do you want to enforce that skipOnError is definetly a boolean? Why then not do `!!settings.skipOnError'?

index.js
gurgunday: Let me test actually

index.js
gurgunday: I did this `!!settings.skipOnError` first, but all other options are type checked, so defaulted to that, just small things

Uzlopak: Oh, if it is for consistency, I have no issue with that.

index.js
gurgunday: Anyway if Fastify is all for speed, like you said, it shouldn't have typeof checks in request handlers

The types are there for a reason

mcollina (approved):
lgtm

Uzlopak:
@mcollina 

Are you sure? Is noop not faster than optional chaining?

gurgunday:
I actually forgot about this one, but the main thing is `onExceeding` and `onExceeded` both had `typeof ===` checks while `onBan` called an empty function, which you say is better, which I don't object to

The PR is simply one about consistency, they should either all have empty default functions or be optionally called — like how either all options should be type checked or none

PS: I think the difference is extremely small

Uzlopak:
What would be more interesting would be a benchmark

gurgunday:
Let me try that even though it's likely that the difference will be hard to measure, especially between an empty function call and optional chaining

gurgunday:
No medians, no statistics, just the highest scores I could get after running each of them 10 times. Obviously not a breakthrough improvement, and the difference between the last 2 is definitely non-significant IMO

My intention here was to have the same way of handling defaults within the codebase

master, node 20
```shell
150 connections


┌─────────┬──────┬──────┬───────┬──────┬────────┬─────────┬────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg    │ Stdev   │ Max    │
├─────────┼──────┼──────┼───────┼──────┼────────┼─────────┼────────┤
│ Latency │ 2 ms │ 2 ms │ 3 ms  │ 5 ms │ 2.2 ms │ 1.57 ms │ 224 ms │
└─────────┴──────┴──────┴───────┴──────┴────────┴─────────┴────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬──────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg      │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┼─────────┤
│ Req/Sec   │ 47839   │ 47839   │ 51551   │ 51903   │ 51073.46 │ 1245.38 │ 47825   │
├───────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┼─────────┤
│ Bytes/Sec │ 16.8 MB │ 16.8 MB │ 18.7 MB │ 18.8 MB │ 18.5 MB  │ 590 kB  │ 16.8 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴──────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.
# of samples: 11

5000 2xx responses, 556809 non 2xx responses
562k requests in 11.01s, 203 MB read
```

PR, node 20
```shell
150 connections


┌─────────┬──────┬──────┬───────┬──────┬─────────┬─────────┬────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max    │
├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼────────┤
│ Latency │ 2 ms │ 2 ms │ 3 ms  │ 5 ms │ 2.16 ms │ 2.12 ms │ 269 ms │
└─────────┴──────┴──────┴───────┴──────┴─────────┴─────────┴────────┘
┌───────────┬───────┬───────┬─────────┬─────────┬──────────┬─────────┬───────┐
│ Stat      │ 1%    │ 2.5%  │ 50%     │ 97.5%   │ Avg      │ Stdev   │ Min   │
├───────────┼───────┼───────┼─────────┼─────────┼──────────┼─────────┼───────┤
│ Req/Sec   │ 48351 │ 48351 │ 52543   │ 52959   │ 51780.37 │ 1455.55 │ 48325 │
├───────────┼───────┼───────┼─────────┼─────────┼──────────┼─────────┼───────┤
│ Bytes/Sec │ 17 MB │ 17 MB │ 19.1 MB │ 19.2 MB │ 18.7 MB  │ 659 kB  │ 17 MB │
└───────────┴───────┴───────┴─────────┴─────────┴──────────┴─────────┴───────┘

Req/Bytes counts sampled once per second.
# of samples: 11

5000 2xx responses, 564578 non 2xx responses
570k requests in 11.02s, 206 MB read
```

PR with empty default functions, node 20
```shell
150 connections


┌─────────┬──────┬──────┬───────┬──────┬─────────┬─────────┬────────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max    │
├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼────────┤
│ Latency │ 2 ms │ 2 ms │ 3 ms  │ 5 ms │ 2.17 ms │ 2.13 ms │ 273 ms │
└─────────┴──────┴──────┴───────┴──────┴─────────┴─────────┴────────┘
┌───────────┬─────────┬─────────┬───────┬─────────┬──────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%   │ 97.5%   │ Avg      │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼───────┼─────────┼──────────┼─────────┼─────────┤
│ Req/Sec   │ 47903   │ 47903   │ 52351 │ 52703   │ 51640.73 │ 1499.95 │ 47898   │
├───────────┼─────────┼─────────┼───────┼─────────┼──────────┼─────────┼─────────┤
│ Bytes/Sec │ 16.8 MB │ 16.8 MB │ 19 MB │ 19.1 MB │ 18.7 MB  │ 679 kB  │ 16.8 MB │
└───────────┴─────────┴─────────┴───────┴─────────┴──────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.
# of samples: 11

5000 2xx responses, 563085 non 2xx responses
568k requests in 11.02s, 206 MB read
```


Code
```js
import Fastify from 'fastify'
import rateLimit from '../index.js'

const fastify = Fastify()
await fastify.register(rateLimit, { max: 5000, timeWindow: '5h', ban: 1 })

fastify.get('/', async (req, reply) => 'hello!')

fastify.listen({ host: '127.0.0.1', port: 5001 }, (err, address) => {
  if (err) throw err
  console.log(address)
})
```

Uzlopak (approved):

PR #320, opened 2023-10-20
use toad-cache, fix bugs, simplify the code

gurgunday:
Rate limiting is a crucial part of a web application, so we must constantly strive to improve `fastify-rate-limit`'s performance and architecture

Before doing some more significant improvements, I wanted to simplify the codebase so that future changes don't get mixed up with these

This PR results in some performance gain thanks to the following changes:

### use `toad-cache`
Toad cache is faster than `tiny-lru` in [important](https://github.com/kibertoad/nodejs-benchmark-tournament/blob/master/cache-get-inmemory/_results/results.md) [benchmarks](https://github.com/kibertoad/nodejs-benchmark-tournament/blob/master/cache-set-inmemory/_results/results.md) and I've shared my interest with kibertoad on how we can centralize all LRU and FIFO packages to his within Fastify

### use our own TTL
I realized while looking at the source code that we already store an `iterationStartTime` using `Date.now()`, we are however not doing enough with this information and just returning it back to user. We can increase performance by disabling the TTL in the LRU (since it would otherwise have more `Date.now` calls) and reseting a request lazily ourselves

### move objects and functions out of the main function
Since `rate-limit` supports being registered multiple times, every time a user does that, we are risking recreating the objects that are inside the main function, which can lead to increased memory usage

### change logic order
In places like the onRequest handler, we would do calculations before checking for conditions that would get us out of the logic flow, for instance we would parse the timeWindow string using `ms` before checking if the hook had already run

### bug fixes
- A flaky test (that would occasionally fail on `master` as well) was fixed
- The route config's `allowList` would get ignored if the global `allowList` property was a function
- when the decorated rateLimit function was used with an option object that didn't have the `timeWindow` property, ex: `fastify.rateLimit({max: 5})`, the options would not get registered

I've added tests for the bug fixes

Benchmark (new vs. old, highest of 5 runs):
```shell
10 connections


┌─────────┬──────┬──────┬───────┬──────┬─────────┬─────────┬───────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max   │
├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼───────┤
│ Latency │ 0 ms │ 0 ms │ 0 ms  │ 0 ms │ 0.01 ms │ 0.04 ms │ 10 ms │
└─────────┴──────┴──────┴───────┴──────┴─────────┴─────────┴───────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬──────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg      │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┼─────────┤
│ Req/Sec   │ 48991   │ 48991   │ 50591   │ 66495   │ 54250.19 │ 6412.41 │ 48982   │
├───────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┼─────────┤
│ Bytes/Sec │ 17.9 MB │ 17.9 MB │ 18.9 MB │ 19.9 MB │ 18.9 MB  │ 517 kB  │ 17.8 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴──────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.
# of samples: 11

150000 2xx responses, 446750 non 2xx responses
597k requests in 11.01s, 208 MB read
```

```shell
10 connections


┌─────────┬──────┬──────┬───────┬──────┬─────────┬─────────┬──────┐
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max  │
├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼──────┤
│ Latency │ 0 ms │ 0 ms │ 0 ms  │ 0 ms │ 0.01 ms │ 0.04 ms │ 7 ms │
└─────────┴──────┴──────┴───────┴──────┴─────────┴─────────┴──────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬──────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg      │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┼─────────┤
│ Req/Sec   │ 47871   │ 47871   │ 49983   │ 66111   │ 53460.37 │ 6605.44 │ 47843   │
├───────────┼─────────┼─────────┼─────────┼─────────┼──────────┼─────────┼─────────┤
│ Bytes/Sec │ 17.7 MB │ 17.7 MB │ 18.7 MB │ 19.7 MB │ 18.6 MB  │ 542 kB  │ 17.7 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴──────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.
# of samples: 11

150000 2xx responses, 438058 non 2xx responses
588k requests in 11.01s, 205 MB read
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,15 +105,15 @@ async function fastifyRateLimit (fastify, settings) {
 
   if (!fastify.hasDecorator('rateLimit')) {
     fastify.decorate('rateLimit', (options) => {
-      const params = options ? mergeParams(globalParams, options) : globalParams
 
-      if (params.timeWindow && params.timeWindow !== globalParams.timeWindow) {
+      if (typeof options === 'object') {
         const newPluginComponent = Object.create(pluginComponent)
-        newPluginComponent.store = newPluginComponent.store.child(Object.assign({}, params))
-        return rateLimitRequestHandler(newPluginComponent, params)
+        const mergedRateLimitParams = mergeParams(globalParams, options)
+        newPluginComponent.store = newPluginComponent.store.child(mergedRateLimitParams)
+        return rateLimitRequestHandler(newPluginComponent, mergedRateLimitParams)
       }
 
-      return rateLimitRequestHandler(pluginComponent, params)
+      return rateLimitRequestHandler(pluginComponent, globalParams)
     })
   }
 
--- a/index.js
+++ b/index.js
@@ -158,6 +158,7 @@ function addRouteRateHook (pluginComponent, globalParams, routeOptions) {
 }
 
 function rateLimitRequestHandler (pluginComponent, params) {
+  const store = pluginComponent.store
   return async function onRequestRateLimiter (req, res) {
     const rateLimitRan = pluginComponent.rateLimitRan
 
@@ -189,7 +190,7 @@ function rateLimitRequestHandler (pluginComponent, params) {
     // We increment the rate limit for the current request
     try {
       const res = await new Promise(function (resolve, reject) {
-        pluginComponent.store.incr(key, (err, res) => {
+        store.incr(key, (err, res) => {
           if (err) {
             reject(err)
             return
--- a/index.js
+++ b/index.js
@@ -188,15 +188,11 @@ function rateLimitRequestHandler (pluginComponent, params) {
     try {
       const res = await new Promise((resolve, reject) => {
         store.incr(key, (err, res) => {
-          if (err) {
-            reject(err)
-            return
-          }
-          resolve(res)
+          err ? reject(err) : resolve(res)
         }, max)
       })
 
-      current = res.current
+      current = res.count
       ttl = res.ttl
     } catch (err) {
       if (!params.skipOnError) {
--- a/index.js
+++ b/index.js
@@ -64,9 +64,9 @@ async function fastifyRateLimit (fastify, settings) {
   globalParams.hook = settings.hook || defaultHook
   globalParams.allowList = settings.allowList || settings.whitelist || null
   globalParams.ban = settings.ban || null
-  globalParams.onBanReach = typeof settings.onBanReach === 'function' ? settings.onBanReach : undefined
-  globalParams.onExceeding = typeof settings.onExceeding === 'function' ? settings.onExceeding : undefined
-  globalParams.onExceeded = typeof settings.onExceeded === 'function' ? settings.onExceeded : undefined
+  globalParams.onBanReach = typeof settings.onBanReach === 'function' ? settings.onBanReach : null
+  globalParams.onExceeding = typeof settings.onExceeding === 'function' ? settings.onExceeding : null
+  globalParams.onExceeded = typeof settings.onExceeded === 'function' ? settings.onExceeded : null
   globalParams.continueExceeding = typeof settings.continueExceeding === 'boolean' ? settings.continueExceeding : false
 
   const rateLimitRan = Symbol('fastify.request.rateLimitRan')
--- a/index.js
+++ b/index.js
@@ -59,7 +59,7 @@ async function fastifyRateLimit (fastify, settings) {
     : typeof settings.timeWindow === 'number' && !isNaN(settings.timeWindow)
       ? settings.timeWindow
       : 1000 * 60
-  globalParams.timeWindowInSeconds = (globalParams.timeWindow / 1000) | 0
+  globalParams.timeWindowInSeconds = Math.floor(globalParams.timeWindow / 1000)
 
   globalParams.hook = settings.hook || defaultHook
   globalParams.allowList = settings.allowList || settings.whitelist || null
@@ -140,7 +140,7 @@ function mergeParams (params1, params2) {
     result.timeWindow = ms.parse(result.timeWindow)
   }
   if (typeof result.timeWindow === 'number' && !isNaN(result.timeWindow)) {
-    result.timeWindowInSeconds = (result.timeWindow / 1000) | 0
+    result.timeWindowInSeconds = Math.floor(result.timeWindow / 1000)
   }
   return result
 }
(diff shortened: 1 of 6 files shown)


gurgunday:
Yeah, should've tested Resis as well, will take a second

Uzlopak:
@kibertoad 
Do you agree with using `toad-cache`. I know the maintainer of toad-cache, and this guy seems shady.

kibertoad:
@Uzlopak I only trust him as far as I can throw him, so no. We need more respectable members of the community in that team for increased confidence, so hopefully we can get some.

(sent invites to you and @gurgunday just in case you are open to it, feel free to decline if not)

Uzlopak:
:D

Thank you

test/not-found-handler-rate-limited.test.js
```diff
@@ -82,21 +82,21 @@ test('Set not found handler can be rate limited with specific options', async t
   t.equal(res.statusCode, 404)
   t.equal(res.headers['x-ratelimit-limit'], '4')
   t.equal(res.headers['x-ratelimit-remaining'], '1')
-  t.equal(res.headers['x-ratelimit-reset'], '1')
+  t.ok(['0', '1', '2'].includes(res.headers['x-ratelimit-reset']))
```
Uzlopak, line 85: Why are the values not set to a fixed value?

gurgunday: That is the flaky test I mentioned above, it fails sometimes on master too if there is enough time for the ttl to go down and get floored to 0 instead of 1

The [first check](https://github.com/fastify/fastify-rate-limit/blob/b867b5463e5bd342fa70687cb82fc95e2c29e6a2/test/not-found-handler-rate-limited.test.js#L79) of that test takes this into account, but the rest of the checks don't, so it would fail from time to time

index.js
```diff
@@ -8,92 +8,71 @@ const RedisStore = require('./store/RedisStore')
...
-  // define the global Time Window
+  // Global time window
   globalParams.timeWindow = typeof settings.timeWindow === 'string'
     ? ms.parse(settings.timeWindow)
     : typeof settings.timeWindow === 'number' && !isNaN(settings.timeWindow)
       ? settings.timeWindow
       : 1000 * 60
-
-  globalParams.timeWindowInSeconds = (globalParams.timeWindow / 1000) | 0
+  globalParams.timeWindowInSeconds = Math.floor(globalParams.timeWindow / 1000)
```
Uzlopak, line 62: Is there a chance to get a number bigger than 2^32?

test/not-found-handler-rate-limited.test.js
Uzlopak: hmm. but it should be then never 2?

index.js
```diff
@@ -8,92 +8,71 @@ const RedisStore = require('./store/RedisStore')
...
 
   globalParams.hook = settings.hook || defaultHook
   globalParams.allowList = settings.allowList || settings.whitelist || null
-  globalParams.ban = settings.ban || null
-  globalParams.onBanReach = typeof settings.onBanReach === 'function' ? settings.onBanReach : undefined
-  globalParams.onExceeding = typeof settings.onExceeding === 'function' ? settings.onExceeding : undefined
-  globalParams.onExceeded = typeof settings.onExceeded === 'function' ? settings.onExceeded : undefined
+  globalParams.ban = settings.ban || 0
```
Uzlopak, line 66: I wonder what this means. 

If i set ban to 1 it means one 429 then you get banned. if i set it to 0 it means you get immediatly banned without seeing a 429. Is there a unit test for that?

index.js
```diff
@@ -228,51 +200,57 @@ function rateLimitRequestHandler (params, pluginComponent) {
...
       respCtx.ban = true
-      params.onBanReach?.(req, key)
+      params.onBanReach(req, key)
     }
 
     throw params.errorResponseBuilder(req, respCtx)
   }
 }
 
+function defaultKeyGenerator (req) { return req.ip }
```
Uzlopak, line 244: before it was an anonymous function having not a this scope (maybe little bit faster)

test/not-found-handler-rate-limited.test.js
gurgunday: Rarely it *is* 2 for the second test as well, the idea there is that it should be within that interval

index.js
gurgunday: Good question, since it is a user parameter and Number is up to 2^64... it could be IMO

index.js
gurgunday: I was just coercing the type to Number since 0 is falsy and equivalent to `null`, `undefined`, or `false`

Ban only works if it is [truthy](https://github.com/fastify/fastify-rate-limit/blob/master/index.js#L253) (!==0), so 0 kind of turns it off in the same way `undefined` does

I find it more consistent to use fasly values and keep a variable type always the same when possible, but you raised a good point that we might in the future want to add a functionality where ban === 0 means a request is always banned, I will revert it

Uzlopak: I dont know if the current implementation did even take care of that.

index.js
gurgunday: I actually think about this more than necessary and am of the same opinion even though I haven't throughly tested it 😂

As you know normal functions have 2 extra objects, `this` and `arguments`, so in theory arrow functions don't have this baggage and if `this` is not of concern, it could mean better

But the engine probably optimizes it when it doesn't see the use of this inside a function

jsumners (commented):
This PR does too many things in one go. From the title, it seems like it should be 3 separate PRs.

gurgunday:
Alright, I’ll break them up 🙏

But I think the refactor is important and should land one way or another

PR #322, opened 2023-10-22
fix RedisStore resetting the TTL unnecessarily and align its behavior with LocalStore

### LINKED ISSUE #304: continueExceeding option does not work as expected
### Prerequisites

### Fastify version

4.18.0

### Plugin version

8.0.1

### Node.js version

18.16.1

### Operating system

Linux

Debian GNU/Linux 11 (bullseye)

### Description

The problem appears to be caused by a bug in the redis implementation of the `@fastify/rate-limit` plugin. When the option `continueExceeding=true`, the operation does not work as expected. Some users were experiencing throttling even for legitimate traffic (below 20 requests/minute - so below the 80 threshold in my case).

There is probably a BUG here: fastify-rate-limit/store/RedisStore.js

```js
'use strict'

const noop = () => { }

function RedisStore (redis, key, timeWindow, continueExceeding) {
  this.redis = redis
  this.timeWindow = timeWindow
  this.key = key
}

RedisStore.prototype.incr = function (ip, cb) {
  const key = this.key + ip
  if (this.continueExceeding) {
    this.redis.pipeline()
      .incr(key)
      .pexpire(key, this.timeWindow)
      .exec((err, result) => {
        if (err) return cb(err, { current: 0 })
        if (result[0][0]) return cb(result[0][0], { current: 0 })
        cb(null, { current: result[0][1], ttl: this.timeWindow })
      })
  } else {
    this.redis.pipeline()
      .incr(key)
      .pttl(key)
      .exec((err, result) => {
        /**
         * result[0] => incr response: [0]: error, [1]: new incr value
         * result[1] => pttl response: [0]: error, [1]: ttl remaining
         */
        if (err) return cb(err, { current: 0 })
        if (result[0][0]) return cb(result[0][0], { current: 0 })
        if (result[1][1] === -1) {
          this.redis.pexpire(key, this.timeWindow, noop)
          result[1][1] = this.timeWindow
        }
        cb(null, { current: result[0][1], ttl: result[1][1] })
      })
  }
}

  child.continueExceeding = routeOptions.continueExceeding
  return child
}

module.exports = RedisStore
```

### Steps to Reproduce

```js
import Fastify from 'fastify'

(async () => {
  const fastify = Fastify()
    max: 80,
    timeWindow: '1 minute',
    continueExceeding: true
  })

  fastify.get('/', (_request, reply) => {
    reply.send({ hello: 'fastify' })
  })

  fastify.listen({ port: 8000 }, (err, address) => {
    if (err) throw err
    console.log('Server running on port 80')
  })
})()
```

### Expected Behavior

When `continueExceeding=true`, renew user limitation when user sends a request to the server when still limited

### LINKED ISSUE #321: Align the behaviors of the LocalStore and RedisStore
### Prerequisites

- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I suspect they don't act the same way currently, especially when `continueExceeding` [is](https://github.com/fastify/fastify-rate-limit/blob/master/store/LocalStore.js) [enabled](https://github.com/fastify/fastify-rate-limit/blob/master/store/RedisStore.js)

Also, since a second issue for this is not necessary, I don't get the logic behind the following:
https://github.com/fastify/fastify-rate-limit/blob/b867b5463e5bd342fa70687cb82fc95e2c29e6a2/index.js#L247-L251

Why do we not return the `timeLeft` like we do with rateReset but the static `timeWindowInSeconds` property when the draft headers are not enabled? To me, retryAfter should simply mimic rateReset

These might need investigating

gurgunday:
Fixes #304 and closes #321

Motivation: **the implementations of the LocalStore and the RedisStore are doing different things**, which is inconsistent and confusing users.

For instance, when `continueExceeding` is `true`, RedisStore *always* resets the TTL of an IP, even when it hasn't yet exceeded the limit. This causes people get rate limited more than necessary because for every request that hasn't exceeded the max, we keep on resetting the ttl instead of letting it go down, while also incrementing the counter.

This PR makes RedisStore mirror LocalStore's behavior. We increment the counter but do not reset the TLL unless the limit is exceeded, after which the TTL is reset for every request as long as the key is alive.

This should land after #320

**If you only want to see the changes of this PR, you can see my local PR: https://github.com/gurgunday/fastify-rate-limit/pull/2**

I still don't know how to create a PR on another PR 🤠

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,15 +105,15 @@ async function fastifyRateLimit (fastify, settings) {
 
   if (!fastify.hasDecorator('rateLimit')) {
     fastify.decorate('rateLimit', (options) => {
-      const params = options ? mergeParams(globalParams, options) : globalParams
 
-      if (params.timeWindow && params.timeWindow !== globalParams.timeWindow) {
+      if (typeof options === 'object') {
         const newPluginComponent = Object.create(pluginComponent)
-        newPluginComponent.store = newPluginComponent.store.child(Object.assign({}, params))
-        return rateLimitRequestHandler(newPluginComponent, params)
+        const mergedRateLimitParams = mergeParams(globalParams, options)
+        newPluginComponent.store = newPluginComponent.store.child(mergedRateLimitParams)
+        return rateLimitRequestHandler(newPluginComponent, mergedRateLimitParams)
       }
 
-      return rateLimitRequestHandler(pluginComponent, params)
+      return rateLimitRequestHandler(pluginComponent, globalParams)
     })
   }
 
--- a/index.js
+++ b/index.js
@@ -158,6 +158,7 @@ function addRouteRateHook (pluginComponent, globalParams, routeOptions) {
 }
 
 function rateLimitRequestHandler (pluginComponent, params) {
+  const store = pluginComponent.store
   return async function onRequestRateLimiter (req, res) {
     const rateLimitRan = pluginComponent.rateLimitRan
 
@@ -189,7 +190,7 @@ function rateLimitRequestHandler (pluginComponent, params) {
     // We increment the rate limit for the current request
     try {
       const res = await new Promise(function (resolve, reject) {
-        pluginComponent.store.incr(key, (err, res) => {
+        store.incr(key, (err, res) => {
           if (err) {
             reject(err)
             return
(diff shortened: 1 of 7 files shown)


gurgunday:
I apologize for the mess, this is an important bug fix that is bundled with a moderately big refactor of the plugin

I am confident the code is better than ever but the confidence comes from being on the shoulders of 2K+ unit tests that were written by our amazing contributors

mcollina (approved):
lgtm

mcollina:
Why?

gurgunday:
I was asked to separate the PRs to make it easier for the reviewers, I will try to introduce the changes one by one

PR #323, opened 2023-10-24
Align stores

### LINKED ISSUE #304: continueExceeding option does not work as expected

### LINKED ISSUE #321: Align the behaviors of the LocalStore and RedisStore

gurgunday:
Fixes #304 and closes #321

Motivation: **the implementations of the LocalStore and the RedisStore are doing different things**, which is inconsistent and confusing users.

For instance, when `continueExceeding` is `true`, RedisStore *always* resets the TTL of an IP, even when it hasn't yet exceeded the limit. This causes people get rate limited more than necessary because for every request that hasn't exceeded the max, we keep on resetting the ttl instead of letting it go down, while also incrementing the counter.

This PR makes RedisStore mirror LocalStore's behavior. We increment the counter but do not reset the TLL unless the limit is exceeded, after which the TTL is reset for every request as long as the key is alive.


package.json
```diff
@@ -42,7 +42,7 @@
   "dependencies": {
     "@lukeed/ms": "^2.0.1",
     "fastify-plugin": "^4.0.0",
-    "tiny-lru": "^11.0.0"
```
jsumners, line 45: This is a breaking change.

index.js
```diff
@@ -83,9 +83,9 @@ async function fastifyRateLimit (fastify, settings) {
     pluginComponent.store = new Store(globalParams)
   } else {
     if (settings.redis) {
-      pluginComponent.store = new RedisStore(settings.redis, settings.nameSpace || 'fastify-rate-limit-', globalParams.timeWindow, settings.continueExceeding)
+      pluginComponent.store = new RedisStore(settings.redis, globalParams.timeWindow, settings.continueExceeding, settings.nameSpace || 'fastify-rate-limit-')
     } else {
-      pluginComponent.store = new LocalStore(globalParams.timeWindow, settings.cache, fastify, settings.continueExceeding)
+      pluginComponent.store = new LocalStore(settings.cache, globalParams.timeWindow, settings.continueExceeding)
```
jsumners, line 88: How does changing these function signatures affect the expectation of user provided `Store` functions?

gurgunday: I doesn't because we check for it in the preceding lines and simply pass `globalParams` to a custom Store, LocalStore is ours (in the sense that we implement it), and RedisStore has to be an `ioredis` instance [1]

https://github.com/fastify/fastify-rate-limit/pull/323/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R81-R84

[1] — We mention it in the README as well

> redis: by default this plugins uses an in-memory store, which is fast but if you application works on more than one server it is useless, since the data is stored locally.
You can pass a Redis client here and magically the issue is solved. To achieve the maximum speed, this plugin requires the use of [ioredis](https://github.com/luin/ioredis). Note: the [default parameters](https://github.com/luin/ioredis/blob/v4.16.0/API.md#new_Redis_new) of a redis connection are not the fastest to provide a rate-limit. We suggest to customize the connectTimeout and maxRetriesPerRequest as in the [example](https://github.com/fastify/fastify-rate-limit/tree/master/example/example.js).

package.json
gurgunday: Why would that be? I have the same explanation as below, the LocalStore is our implementation that cannot be overridden – it behaves the same

Uzlopak:
@gurgunday 

You hav some merge conflicts. 

To be honest i dont like the ['0', '1', '2'].includes assertion. Before it was maybe flaky, and we maybe need to run the tests sequentially. But it would give use feedback on how reliant the rate-limitting is. Now it can be potentially unreliably and we could be blind to that.

gurgunday:
Here are a few things we can do in terms of tests:

The first check of Retry-After nearly always returns the timeWindow, but the rest, there just isn't a single value that we can pinpoint

If timeWindow is 3000, the first call is fine, 3000, but the second can return 2700, which will be floored to 2, and sometimes there are 800ms clock ticks too after which could it could take 210ms to create another Date and now it would be floored to 0

I was thinking, before we even fix these, that Math.ceil might be more appropriate than .floor in terms of rateLimit timers

Because really a client shouldn't see a TTL of 600ms floored to 0 seconds, but they *should* see 1

Uzlopak:
Floor is obviously more restricitive than Ceil.

Uzlopak (approved):
LGTM

PR #324, opened 2023-10-24
Retry-After: return the TTL instead of timeWindow

gurgunday:
`Retry-After`, the internet standard counterpart of `x-ratelimit-reset`, was for some reason always being set to the `timeWindow` property when rate limiting took place - so for instance even when a user had to wait for 5 more seconds for rate limiting to be lifted (with a timeWindow of 15 seconds), the user would still see on the header the number 15 instead of 5

[I traced back a change](https://github.com/fastify/fastify-rate-limit/commit/6de4257297e7bb23d6fc4d0e1eeefa36d3b987a5#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R188) that, while introducing the optional draft headers feature, had done away with this but probably couldn't change the default behavior to make it non-breaking

[While another breaking PR is up](https://github.com/fastify/fastify-rate-limit/pull/323), I think we can also make this change


jsumners (approved):
Looks good to me.

PR #325, opened 2023-10-24
Use `Math.ceil` for timeLeft calculation

gurgunday:
Last (small) breaking change, promise 😁

Here's the case for it: it changes nothing on our side, but sending an endpoint `Retry-After: 0` is arguably worse than sending `Retry-After: 1` when they still have 900ms of blockage

Also, the whole idea of sending `Retry-After: 0` makes no sense

And finally, all tests that contained checks against multiple values have been modified

### DIFF
--- a/index.js
+++ b/index.js
@@ -228,7 +228,7 @@ function rateLimitRequestHandler (params, pluginComponent) {
       }
     }
 
-    const timeLeft = Math.floor(ttl / 1000)
+    const timeLeft = Math.ceil(ttl / 1000)
 
     if (current <= maximum) {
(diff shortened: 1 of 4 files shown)


jsumners (approved):
Looks good to me.

Uzlopak:
I really had to think about it and I think it is the correct way.

Imagining following case:

You have 1.9 seconds left. Math.floor would return 1 and means in 1 second you can make your next request. You make your next request in 1 second and get 0.9 as the internal value and get again a 429 error. But with Math.ceil we get for 1.9 seconds the value 2 and you would do the next request in 2 seconds and dont get the 429 error.

Uzlopak (approved):

PR #327, opened 2023-10-25
feat: autoPipelining

gurgunday:
Not eager to explain why it's structured this way... I'm open to suggestions

I could do a much better design if we added support for async `incr` functions and made more breaking changes, which I have done in another branch, but one step at a time... 🤣

If anyone finds this messy, I can just remove the old version but we would really have to recommend users to enable `autoPipelining`

### DIFF
--- a/store/RedisStore.js
+++ b/store/RedisStore.js
@@ -33,44 +33,37 @@ RedisStore.prototype.incrAutoPipe = async function (ip, cb, max) {
   cb(null, { current, ttl })
 }
 
-RedisStore.prototype.incrNormal = function (ip, cb, max) {
+RedisStore.prototype.incrNormal = async function (ip, cb, max) {
   const key = this.key + ip
-
-  this.redis.pipeline()
+  const result = await this.redis.pipeline()
     .incr(key)
     .pttl(key)
-    .exec((err, result) => {
-    /**
-     * result[0] => incr response: [0]: error, [1]: new incr value
-     * result[1] => pttl response: [0]: error, [1]: ttl remaining
-     */
-      if (err || result[0][0] || result[1][0]) {
-        cb(err || result[0][0] || result[1][0], null)
-        return
-      }
+    .exec()
 
-      if (result[1][1] === -1) {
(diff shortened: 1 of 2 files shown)


Uzlopak (changes requested):
I add a blocker.

gurgunday:
@Uzlopak what do you think? This is without using Lua and should work, having two incr functions is not great

Uzlopak:
I am currently bisecting node. I will come back when I finished with node

gurgunday (commented):
I don’t find this code pleasing to look at

.pipeline does the same thing without mixing async, await, and callbacks

I’ll let this stay open for a bit and close it

store/RedisStore.js
```diff
@@ -5,46 +5,70 @@ function RedisStore (redis, timeWindow, continueExceeding, key) {
...
 
-RedisStore.prototype.incr = function (ip, cb, max) {
+RedisStore.prototype.incrAutoPipeline = async function (ip, cb, max) {
   const key = this.key + ip
+  const [current, ttl] = await Promise.all([this.redis.incr(key), this.redis.pttl(key)])
+
+  if (ttl === -1) {
+    // Item just got created
+    this.redis.pexpire(key, this.timeWindow).catch()
+    cb(null, { current: 1, ttl: this.timeWindow })
```
mcollina, line 23: I think we need to add these as callback to pexpire to avoid test issues

gurgunday: Yeah I think that’s an improvement, just wanted to keep the behavior the same - it was like this before even with callbacks

gurgunday: What do you think about making the functions async? Without making them async, I can’t group them together without .pipeline

mcollina: that would be a semver-major.

mcollina: +1 for me

Uzlopak: I really would prefer a customCommand which runs the db related stuff on redis over more fiddling in the business code.

PR #328, opened 2023-10-25
RedisStore: don't return an object after error

gurgunday:
this doesn't change anything, it's better to return null because the information will be discarded anyway...


Uzlopak (approved):
makes sense

PR #329, opened 2023-10-26
perf: remove isNaN check from the handler

gurgunday:
We don't need to call `isNaN` for each request, we can just check for it once during route registration


gurgunday:
Coverage

climba03003 (approved):

PR #330, opened 2023-10-26
Move Redis logic to Redis-Side

gurgunday:
@Uzlopak asks, we implement ✊

The obvious benefit is that we now make 1 request per function call

Not sure what I think about using Lua as a JS enjoyer though — arrays are 1-indexed 🤢


gurgunday:
I need opinions before making this change 🙏

climba03003 (approved):
LGTM. but I recommend not to eagerly merge, wait for more people to check.

gurgunday:
Found a nasty bug

store/RedisStore.js
```diff
@@ -1,50 +1,88 @@
...
+    this.redis.defineCommand('rateLimitCE', {
+      numberOfKeys: 1,
+      lua: luaContinueExceeding
+    })
+  } else if (!this.continueExceeding && !this.redis.rateLimit) {
+    this.redis.defineCommand('rateLimit', {
+      numberOfKeys: 1,
+      lua: luaBasic
+    })
+  }
```
climba03003, line 67:
```suggestion
  this.redis = redis
  this.timeWindow = timeWindow
  // maybe undfined, but would be fine
  const oldContinueExceeding = this.continueExceeding
  this.continueExceeding = continueExceeding
  this.key = key
  
  if (!this.redis.rateLimit || oldContinueExceeding !== this.continueExceeding) {
    // we need to define rate limit command if
    // 1. not exist before
    // 2. continue exceeding value is changed
    this.redis.defineCommand('rateLimit', {
      numberOfKeys: 1,
      lua: this.continueExceeding ? luaContinueExceeding : luaBasic
    })
  }
```

climba03003: I don't know if redis allowed to redefine the command?

gurgunday: Yeah that's what came to my mind at first, but here's why that won't work: redis is an object, and as you know when we redefine the `rateLimit` property, parent routes' `this.redis.rateLimit` will be updated as well, because they share the same object reference

This will essentially break any global config that sets `continueExceeding: false` if the last registered route has a config that enables continueExceeding or vice versa

store/RedisStore.js
```diff
@@ -1,50 +1,88 @@
...
+
+  -- If the key is new then set its TTL
+  if ttl == -1 then
+      redis.call('PEXPIRE', key, timeWindow)
+      return {1, timeWindow}
+  -- If the key's incremented value has exceeded the max value, then reset its TTL
+  elseif value > max then
+      redis.call('PEXPIRE', key, timeWindow)
+      return {value, timeWindow}
+  end
```
climba03003, line 46:
```suggestion
  elseif value > max then
      redis.call('PEXPIRE', key, timeWindow)
      return {value, timeWindow}
  end
```

If this part is the only difference, why don't we share the same script with another argument?

gurgunday: You are right, this whole complexity is caused by having two scripts in the first place

I will update

climba03003 (approved):
Still LGTM. much simpler in JavaScript

Uzlopak:
Please add an example example-auto-pipeline.js (copy of example.js) with autoPipelining set to true.

Add documentation regarding autoPipelining.

gurgunday:
None of the examples are working anymore btw, they need to await `fastify.register`

Uzlopak:
@gurgunday 
Thank you. I was currently testing your PR and was wondering why the rate-limitting is not working.

climba03003:
> Thank you. I was currently testing your PR and was wondering why the rate-limitting is not working.

Obvious reason, using `onRoute` hook.

Uzlopak:
Yeah..

airhorns:
Just chiming in from afar, I think this is a major improvement, in my experience with redis in the past executing in scripts is far far faster than roundtripping for each element. The only downsides are you need a bit of complexity in the client to load the script if it hasn't been loaded already (which ioredis handles just fine), and you need to use a redis version with lua support, and it's been out for a long time. +1 from me on the direction of this PR. Do the existing tests cover this already?

gurgunday:
@airhorns hey, thanks a lot for the comment

We do have tests that check if Redis behavior is correct

GitHub actions by default uses the latest version and I’m not sure if we test for redis versions, but this release will most definitely be a major bump

gurgunday:
Just to add, I’m no Lua expert so I would appreciate more people confirming that the script looks good to them 🙏

Uzlopak:
I dont think that we will get some feedback from redis discord. 

So i am om with this PR.

Uzlopak (approved):
Awesome work.

marco-ippolito (approved):
LGTM

PR #331, opened 2023-10-26
feat: ban users directly when rate limited

gurgunday:
there might be cases where we want to directly ban a visitor when they hit our limits


gurgunday:
Very interesting...

gurgunday:
@Eomm how is this even possible?

This PR hasn't changed anything in those files and they simply started to fail

index.js
```diff
@@ -72,7 +72,7 @@ async function fastifyRateLimit (fastify, settings) {
 
   globalParams.hook = settings.hook || defaultHook
   globalParams.allowList = settings.allowList || settings.whitelist || null
-  globalParams.ban = settings.ban || 0 // 0 means turned off
+  globalParams.ban = typeof settings.ban === 'number' && !isNaN(settings.ban) ? settings.ban : Infinity
```
Uzlopak, line 75:
```suggestion
  globalParams.ban = typeof settings.ban === 'number' && Number.isFinite(settings.ban) && settings.ban >= 0 ? Math.trunc(settings.ban) : Infinity
```

Maybe?

index.js
```diff
@@ -58,21 +59,22 @@ async function fastifyRateLimit (fastify, settings) {
...
     : defaultMax
 
   // Global time window
   globalParams.timeWindow = typeof settings.timeWindow === 'string'
     ? ms.parse(settings.timeWindow)
-    : typeof settings.timeWindow === 'number' && !isNaN(settings.timeWindow)
-      ? settings.timeWindow
-      : 1000 * 60
-  globalParams.timeWindowInSeconds = Math.floor(globalParams.timeWindow / 1000)
+    : typeof settings.timeWindow === 'number' && !Number.isNaN(settings.timeWindow) && settings.timeWindow >= 0
```
Uzlopak, line 69: set timeWindow to Infinity. kaboom?

gurgunday: Nice, everyone gets a certain number of requests and then they're banned for life 🤣

You think that scenario is absurd?

gurgunday: Actually I don't even know if it would work in redis or not so I'm not going to risk it

gurgunday:
Ridiculous, gotta take a break 🤠🤣

index.js
```diff
@@ -58,21 +59,22 @@ async function fastifyRateLimit (fastify, settings) {
...
     ? ms.parse(settings.timeWindow)
-    : typeof settings.timeWindow === 'number' && !isNaN(settings.timeWindow)
-      ? settings.timeWindow
-      : 1000 * 60
-  globalParams.timeWindowInSeconds = Math.floor(globalParams.timeWindow / 1000)
+    : typeof settings.timeWindow === 'number' && Number.isFinite(settings.timeWindow) && settings.timeWindow >= 0
+      ? Math.trunc(settings.timeWindow)
+      : defaultTimeWindow
+
+  globalParams.timeWindowInSeconds = Math.trunc(globalParams.timeWindow / 1000)
```
Uzlopak, line 73: didnt we decide to ceil it?

gurgunday: We decided to ceil it when we are [sending it to the user](https://github.com/fastify/fastify-rate-limit/blob/efcb43cebaed1dbade63c1a3790bc2868f778aa1/index.js#L219C46-L219C46), but I don't know if we should ceil that property as well

I actually wanted to remove `timeWindowInSeconds` because it is used nowhere in the code... I just hesitate to do it because we pass globalParams to custom stores and people might be using it

gurgunday: Should I do ceil there as well? Or just remove it maybe?

Uzlopak: idk you are deeper in the code than I

gurgunday: I would remove it because a CustomStore user can just divide the timeWindow by 1000 themselves within the constructor and do a ceil or trunc (if it ends up being a decimal) depending on what they want to do

But right now, it was `.floor` before, so it's non-breaking in that regard

gurgunday: I actually find it dangerous, because it might cause desync issues when timeWindow is floored or ceild

climba03003 (commented):
No particular preference on this issue.

But before `0` means disable the ban, now it means ban directly.
So it would be a breaking change.

Eomm (commented):
we wrote `ban: 2, // default null` in the docs 😱

README.md
```diff
@@ -114,7 +114,7 @@ You can pass the following options during the plugin registration:
 await fastify.register(import('@fastify/rate-limit'), {
   global : false, // default true
   max: 3, // default 1000
-  ban: 2, // default null
+  ban: 2, // default Infinity
```
mcollina, line 117: This change likely needs a bit more docs than this

gurgunday: Yeah probably types should be changed as well… will take a look at it today

Uzlopak: So wait. Null could mean disabled and 0 could a mean immediatly block. Better than infinity?

README.md
```diff
@@ -140,9 +140,9 @@ await fastify.register(import('@fastify/rate-limit'), {
...
-- `ban`: is the maximum number of 429 responses to return to a single client before returning 403. When the ban limit is exceeded the context field will have `ban=true` in the errorResponseBuilder. This parameter is an in-memory counter and could not work properly in a distributed environment.
+- `global` : indicates if the plugin should apply rate limiting to all routes within the encapsulation scope.
+- `max`: maximum number of requests a single client can perform inside a timeWindow. It can be an async function with the signature `async (request, key) => {}` where `request` is the Fastify request object and `key` is the value generated by the `keyGenerator`. The function **must** return a number.
+- `ban`: maximum number of 429 responses to return to a client before returning 403 responses. When the ban limit is exceeded, the context argument that is passed to `errorResponseBuilder` will have its `ban` property set to `true`. This parameter is an in-memory counter and will not work properly in a distributed environment.  **Note:** `0` can also be passed to directly return 403 responses when a client exceeds the limit.
```
Uzlopak, line 145: regarding the last sentence:

0 is the limit ("exceeds the limit"), and returning 403 response is the effect. I would prefer to write that 0 is a valid value and results in an immediatly ban and that the server will return a 403. 

Maybe @Fdawgs has some suggestions

Uzlopak (approved):
LGTM, but readme should be improved as remarked.

Fdawgs (approved):

README.md
mcollina: I would split this into two configs minumum

Uzlopak: Should he implement it or is this a suggestion?

gurgunday: So something like `ban` and `banCanBe0`?

I think just applying the same logic for `ban` when it's defined as a numeric value (including 0) is simpler

Uzlopak:
@Fdawgs 
any remarks regarding the wording of the readme?

PR #332, opened 2023-10-27
chore(package): explicitly declare js module type

Fdawgs:
[Node 21.1.0 added a flag to detect module types](https://github.com/nodejs/node/releases/tag/v21.1.0), which will probably become default in the future. Declaring the type will cause Node to skip detection on startup, reducing startup time.

Declaring the package type is also considered good practice according to https://nodejs.org/api/modules.html#enabling anyway, and the [main fastify repo already does it](https://github.com/fastify/fastify/blob/9b8a7825dc033887d293549e40284284bf27c5a5/package.json#L6).

Will update the rest of the repos and point to this PR in their PRs.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "8.0.3",
   "description": "A low overhead rate limiter for your routes",
   "main": "index.js",
+  "type": "commonjs",
   "types": "types/index.d.ts",
   "scripts": {
     "lint": "standard",


gurgunday (approved):

Fdawgs:
Not too sure on repos like https://github.com/fastify/fastify-flash that are written in TS but compile to CJS

simoneb:
To be honest I don't think this will ever happen, as it would break existing applications. I would find it more likely that from a major version onwards, it will default it to module instead of cjs.

PR #334, opened 2023-10-30
remove elseif from the lua script

gurgunday:
We are returning in each case so we don't need the `elseif`

Can't believe I realized this after it got merged


Uzlopak (approved):

Eomm:
Does it mean we don't have a test for it? 😏

gurgunday:
Haha no no, we do have tests, this change didn't alter Redis's behavior

There are return statements inside the conditional checks, eliminating the need for an `else if` clause; a return statement would already exit the control flow

PR #336, opened 2023-11-01
Redis should work with ban

### LINKED ISSUE #335: Make ban work with Redis
### Prerequisites

- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

The ban property currently only works in-memory, we can do better

### Motivation

_No response_

### Example

_No response_

gurgunday:
Closes #335

~Now a custom store user should handle when ban happens themselves - I believe this will actually lead to more flexible configurations and simpler code because that's how `max` currently works as well~

~If someone objects to my last point, I can make it non-breaking by adding a simple check and falling back to old behavior if `res.ban` is `undefined`, which means CustomStore didn't do anything with the newly added ban parameter~


Uzlopak:
How is the performance?

gurgunday:
Will check it

gurgunday:
@Uzlopak I reset my Mac and can't set up Redis right now... I don't think it's that bad and this change is no longer breaking

store/LocalStore.js
```diff
@@ -8,31 +8,36 @@ function LocalStore (cache = 5000, timeWindow, continueExceeding) {
   this.continueExceeding = continueExceeding
 }
 
-LocalStore.prototype.incr = function (ip, cb, max) {
+LocalStore.prototype.incr = function (ip, cb, max, ban) {
```
Eomm, line 11: we should add it to the doc about the `CustomStore`

PR #339, opened 2023-11-04
ci: migrate to workflows

gurgunday:


gurgunday:
Why is it not working, @Eomm?

I don't know the current workflow that well

Fdawgs:
@gurgunday Shared workflow tests on macOS, Ubuntu, and Windows. Services like the redis one here only work with Ubuntu.

gurgunday:
I can't use `@3` for this workflow (ci-redis), I had to use `@main` - where can I find the versions?

Uzlopak (approved):

PR #347, opened 2023-12-13
revert: handle ban in store

gurgunday:
Hey everyone,

The change that moved ban handling to the store-side **in a non-breaking manner** hasn't accomplished anything meaningful in terms of fixing the sync issue between multiple deployments when an in-memory counter is used, I basically misunderstood the problem

Naturally there will always be this issue with in-memory counters, and moving the ban calculation to the store just increased the network payload for Redis configurations, albeit a little bit

Now, I'm pretty sure not many people have custom stores that handle `ban` themselves, because the change is pretty new (a month old), but in any case, I will be targeting next to be sure

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,7 +206,6 @@ function rateLimitRequestHandler (pluginComponent, params) {
     let current = 0
     let ttl = 0
     let timeLeftInSeconds = 0
-    let ban = false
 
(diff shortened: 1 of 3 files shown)


gurgunday:
I should target next I guess

mcollina (approved):
lgtm

PR #350, opened 2023-12-13
Revert workflow update

gurgunday:
I don't know why at the moment, but [centralized workflows for redis](https://github.com/fastify/workflows/blob/main/.github/workflows/plugins-ci-redis.yml) are failing

We should revert temporarily to v3


gurgunday:
Pinged to let you know

Fdawgs (approved):

PR #356, opened 2024-01-08
refactor: make RedisStore default key a default parameter

gurgunday:
Small refactor that moves everything related to RedisStore to its own file, and uses the validated `globalParams` properties where possible


mcollina (commented):
lgtm

Issue #283, opened 2023-01-19
Function for `timeWindow`


AlexRMU:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Add a function so that you can do a limit, for example 10 per minute AND 100 per hour AND 1000 per day.

### Example

```js
rateLimit: {
    max: async (req, key) => { /*... */ },
    timeWindow: async (req, key) => { /*... */ },
    // or
    max: {
        // key = window
        // value = max
        [60 * 1000]: 10,
        [60 * 60 * 1000]: async (req, key) => { /*... */ },
    }
}
```

Uzlopak:
Well this feature request raises questions. What are the headers on the response you gonna set? I think it is useful what you suggest, but needs more specs. So please elaborate.

mindrunner:
+1 for this feature :)

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

AlexRMU:
I tried, but realized that it would take a lot of architecture changes.
I can help with tips or a review.

gurgunday:
I think something similar can be achieved by using a function for the max parameter

mindrunner:
https://github.com/fastify/fastify-rate-limit/pull/357

mindrunner:
I added `function` feature for `timeWindow`. Please have a look  :)

gurgunday:
Closed by #357

PR #357, opened 2024-01-12
Add function support for timeWindow

mindrunner:
This update allows `timeWindow` to be a function that can dynamically determine the time window based on the request and key. The code refactor has been applied across both the Redis and Local stores. Tests for this new functionality have been added to ensure correctness. The TypeScript definition was also updated to reflect this change.

https://github.com/fastify/fastify-rate-limit/issues/283

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,11 +64,18 @@ async function fastifyRateLimit (fastify, settings) {
     : defaultMax
 
   // Global time window
-  globalParams.timeWindow = typeof settings.timeWindow === 'string'
-    ? ms.parse(settings.timeWindow)
-    : typeof settings.timeWindow === 'number' && Number.isFinite(settings.timeWindow) && settings.timeWindow >= 0
-      ? Math.trunc(settings.timeWindow)
-      : defaultTimeWindow
+  const twType = typeof settings.timeWindow
+  globalParams.timeWindow = defaultTimeWindow
+  if (twType === 'function') {
+    globalParams.timeWindow = settings.timeWindow
+  } else if (twType === 'string') {
+    globalParams.timeWindow = ms.parse(settings.timeWindow)
+  } else if (
+    twType === 'number' &&
+    Number.isFinite(settings.timeWindow) && settings.timeWindow >= 0
+  ) {
+    globalParams.timeWindow = Math.trunc(settings.timeWindow)
+  }
 
   globalParams.hook = settings.hook || defaultHook
   globalParams.allowList = settings.allowList || settings.whitelist || null
@@ -147,7 +154,7 @@ function mergeParams (...params) {
     result.timeWindow = ms.parse(result.timeWindow)
   } else if (typeof result.timeWindow === 'number' && Number.isFinite(result.timeWindow) && result.timeWindow >= 0) {
     result.timeWindow = Math.trunc(result.timeWindow)
-  } else {
+  } else if (typeof result.timeWindow !== 'function') {
     result.timeWindow = defaultTimeWindow
   }
 
@@ -180,7 +187,7 @@ function addRouteRateHook (pluginComponent, params, routeOptions) {
 
 function rateLimitRequestHandler (pluginComponent, params) {
   const { rateLimitRan, store } = pluginComponent
-  const timeWindowString = ms.format(params.timeWindow, true)
+  const timeWindowString = ms.format(typeof params.timeWindow === 'function' ? params.timeWindow() : params.timeWindow, true)
 
   return async (req, res) => {
     if (req[rateLimitRan]) {
@@ -204,6 +211,7 @@ function rateLimitRequestHandler (pluginComponent, params) {
     }
 
     const max = typeof params.max === 'number' ? params.max : await params.max(req, key)
+    const timeWindow = typeof params.timeWindow === 'number' ? params.timeWindow : await params.timeWindow(req, key)
     let current = 0
     let ttl = 0
     let timeLeftInSeconds = 0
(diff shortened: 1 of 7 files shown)


mcollina (commented):
Could you update the tests for the types?

index.js
```diff
@@ -64,11 +64,13 @@ async function fastifyRateLimit (fastify, settings) {
...
-    : typeof settings.timeWindow === 'number' && Number.isFinite(settings.timeWindow) && settings.timeWindow >= 0
-      ? Math.trunc(settings.timeWindow)
-      : defaultTimeWindow
+  globalParams.timeWindow = typeof settings.timeWindow === 'function'
+    ? settings.timeWindow
+    : typeof settings.timeWindow === 'string'
+      ? ms.parse(settings.timeWindow)
+      : typeof settings.timeWindow === 'number' && Number.isFinite(settings.timeWindow) && settings.timeWindow >= 0
+        ? Math.trunc(settings.timeWindow)
+        : defaultTimeWindow
```
jsumners, line 73: I realize you didn't start this fire, but please put it out:

```suggestion
	const twType = typeof settings.timeWindow
	globalParams.timeWindow = defaultTimeWindow
	if (twType === 'function') {
		globalParams.timeWindow = settings.timeWindow
	} else if (twType === 'string') {
		globalParams.timeWindow = ms.parse(settings.timeWindow)
	} else if (
		twType === 'number' &&
		Number.isFinite(settings.timeWindow) &&
		settings.timeWindow >= 0
	) {
		globalParams.timeWindow = Math.trunc(settings.timeWindow)
	}
```

jsumners (dismissed):

store/LocalStore.js
```diff
@@ -4,21 +4,20 @@ const { LruMap: Lru } = require('toad-cache')
 
 function LocalStore (cache = 5000, timeWindow, continueExceeding) {
   this.lru = new Lru(cache)
-  this.timeWindow = timeWindow
   this.continueExceeding = continueExceeding
 }
 
-LocalStore.prototype.incr = function (ip, cb, max, ban) {
+LocalStore.prototype.incr = function (ip, cb, timeWindow, max, ban) {
```
gurgunday, line 10: This will be a breaking change since you are modifying the signature of the increment method, this method is called with custom store implementations too, so they would need to be modified

Another change that modifies this method's signature is queued up in the `next` branch ([it removed the `ban` parameter from `incr`](https://github.com/fastify/fastify-rate-limit/commit/fe1ca6f2fd7d16f590fb34ca8a48acdc508ad293)), I suggest releasing these changes together in the next major that will support Fastify v5

index.js
```diff
@@ -214,7 +217,7 @@ function rateLimitRequestHandler (pluginComponent, params) {
       const res = await new Promise((resolve, reject) => {
         store.incr(key, (err, res) => {
           err ? reject(err) : resolve(res)
-        }, max, params.ban)
+        }, timeWindow, max, params.ban)
```
gurgunday, line 220: This will be a breaking change since you are modifying the signature of the increment method, this method is called with custom store implementations too, so they would need to be modified

Another change that modifies this method's signature is queued up in the next branch ([it removed the ban parameter from incr](https://github.com/fastify/fastify-rate-limit/commit/fe1ca6f2fd7d16f590fb34ca8a48acdc508ad293)), I suggest releasing these changes together in the next major that will support Fastify v5

mindrunner: Cool, sounds good! Do you have an idea how far in the future fastify 5 might be?

mindrunner:
> Could you update the tests for the types?

done

mcollina (approved):
lgtm

gurgunday:
@mindrunner can you try targeting `next`?

mindrunner:
> @mindrunner can you try targeting `next`?

done

index.js
```diff
@@ -180,7 +187,7 @@ function addRouteRateHook (pluginComponent, params, routeOptions) {
 
 function rateLimitRequestHandler (pluginComponent, params) {
   const { rateLimitRan, store } = pluginComponent
-  const timeWindowString = ms.format(params.timeWindow, true)
+  const timeWindowString = ms.format(typeof params.timeWindow === 'function' ? params.timeWindow() : params.timeWindow, true)
```
gurgunday, line 190: Minor feedback:

Passing no arguments is not a valid signature according to the type information I think, you can make the parameters optional

Also, just for consistency, let's `await` this as well since it can return a Promise

```suggestion
  const timeWindowString = ms.format(typeof params.timeWindow === 'function' ? await params.timeWindow() : params.timeWindow, true)
```

index.js
gurgunday: After March… but we started making the major releases for the plugins, so don’t see any reason not to do it here as well

index.js
mindrunner: That was not optimal. Thanks for spotting. I simplified it a bit. No need for optional parameters. Should be better now.

index.js
mindrunner: Awesome! :)

index.js
gurgunday: Yep, looks great!

Thanks

gurgunday (approved):
Looks good!

index.js
mindrunner: @gurgunday any chance for a realease any time soon?

gurgunday: @mcollina should we release `next` like some other plugins?

We'd no longer be testing for node 14 but it's fine I think

PR #360, opened 2024-01-29
docs: update `timeWindow` description

gurgunday:
followup to https://github.com/fastify/fastify-rate-limit/pull/357

PR #363, opened 2024-02-22
fix: remove unnecessary property `timeWindow` from RedisStore and LocalStore

gurgunday:
We no longer need to have a `timeWindow` property since it will be passed dynamically to the `incr` method


gurgunday:
Note: this change won't affect custom stores since to them we pass all the options anyway

mcollina (changes requested):
Can you update the docs?

gurgunday:
Here: https://github.com/fastify/fastify-rate-limit/pull/360

mcollina (approved):
lgtm

PR #370, opened 2024-07-06
update for v5

gurgunday:
`t.teardown` is problematic, since the tests are sync, we can just clear them up in the end

### DIFF
--- a/.taprc
+++ b/.taprc
@@ -1,9 +1,3 @@
-jobs: 1
-
-branches: 96
-functions: 100
-lines: 100
-statements: 98
-
+disable-coverage: true
 files:
   - test/**/*.test.js
--- a/.taprc
+++ b/.taprc
@@ -1,3 +1,4 @@
+jobs: 1
 disable-coverage: true
 files:
   - test/**/*.test.js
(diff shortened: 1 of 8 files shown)


.taprc
```diff
@@ -1,9 +1,4 @@
 jobs: 1
-
-branches: 96
-functions: 100
-lines: 100
-statements: 98
-
+disable-coverage: true
```
jean-michelet, line 2: Isn't it better to impose a fixed limit, so that we're aware when the code coverage drop?

Maybe we can open an issue to increase code coverage if possible.

gurgunday: > Maybe we can open an issue to increase code coverage if possible.

Yes

> Isn't it better to impose a fixed limit, so that we're aware when the code coverage drop?

I'm not aware if that exists in tap@20, happy to add it back if it exists

jean-michelet: > I'm not aware if that exists in tap@20, happy to add it back if so

Yep, they broke a few things.
Maybe it's better to fix coverage in an independent PR, will take a look tomorrow

Fdawgs (approved):

PR #382, opened 2024-09-26
perf: reduce optional chaining usage

gurgunday:
From the [V8 blog](https://v8.dev/features/optional-chaining):

> Still, be considerate about using more than one optional chaining operator in a single chain. If a value is guaranteed to not be nullish, then using ?. to access properties on it is discouraged.

It adds more bytecode to each property access


dancastillo (approved):
LGTM - nice optimization!

Issue #279, opened 2022-12-15
Time window with exponential backoff


KaranH15:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Right now in timeWindow option we can set retries for regular interval of time. There is no option for allowing exponential backoff time. E.g If I want user to retry for following time windows - 1min, 5 min, 15min ........ No option to configure that rn.

climba03003:
The current implementation do not keep how many times for an `ip` to be banned.
Those data would require a `persistence` storage other than `redis`.

I think the idea good, but not sure if it worth for storing such data.

Uzlopak:
Just because I have the code currently open in my vscode: Exponential back off can be implemented by the following code:

```js
/**
 * Exponential backoff is the process of a client periodically retrying a
 * failed request over an increasing amount of time. It is a standard error
 * handling strategy for network applications. The Core Reporting API is
 * designed with the expectation that clients which choose to retry failed
 * requests do so using exponential backoff. Besides being "required", using
 * exponential backoff increases the efficiency of bandwidth usage, reduces
 * the number of requests required to get a successful response, and
 * maximizes the throughput of requests in concurrent environments.
 *
 * @see https://developers.google.com/analytics/devguides/reporting/core/v3/errors#backoff
 */
function exponentialDelay(retryNumber) {
	const delayInSeconds = Math.pow(2, retryNumber)
	const randomMs = ~~(1000 * Math.random())
	return delayInSeconds * 1000 + randomMs
}
```

aniketcodes:
This is done. Need inputs/suggestions. #387

PR #387, opened 2024-10-28
Added exponential backoff

aniketcodes:
This PR adds the capability to add exponential backoff.
#279 

#### Checklist


Uzlopak (dismissed):
I dont like it somehow. Blocking it so that i can think about it in a free minute.

store/LocalStore.js
```diff
@@ -24,7 +25,13 @@ LocalStore.prototype.incr = function (ip, cb, timeWindow, max) {
     ++current.current
 
     // Reset TLL if max has been exceeded and `continueExceeding` is enabled
-    if (this.continueExceeding && current.current > max) {
+    if (this.exponentialBackoff && current.current > max) {
+      // Handle exponential backoff
+      const backoffExponent = current.current - max - 1
+      const ttl = timeWindow * Math.pow(2, backoffExponent)
```
gurgunday, line 31:
```suggestion
      const ttl = timeWindow * (2 ** backoffExponent)
```

store/RedisStore.js
```diff
@@ -21,12 +27,20 @@ const lua = `
     redis.call('PEXPIRE', key, timeWindow)
     ttl = timeWindow
   end
+  
+  -- If the key is new or if its incremented value has exceeded the max value and exponential backoff is enabled then set its TTL
+  if ttl == -1 or (exponentialBackoff and current > max) then
+    local backoffExponent = current - max - 1
+    ttl = math.min(timeWindow * math.pow(2, backoffExponent), MAX_SAFE_INTEGER)
```
gurgunday, line 34: Same here, math.pow is deprecated in Lua I believe:

```suggestion
    ttl = math.min(timeWindow * (2 ^ backoffExponent), MAX_SAFE_INTEGER)
```

store/LocalStore.js
```diff
@@ -24,7 +25,13 @@ LocalStore.prototype.incr = function (ip, cb, timeWindow, max) {
     ++current.current
 
     // Reset TLL if max has been exceeded and `continueExceeding` is enabled
```
gurgunday, line 27: This comment is about continue exceeding, please keep the original order here:

it should be:
```
  if (this.continueExceeding)
  else if (this.exponentialBackoff)
  else
```

gurgunday (commented):
LGTM other than a final nit

store/LocalStore.js
aniketcodes: Done

aniketcodes:
Hi @gurgunday , @Uzlopak I have updated the lua script to make the ttl resetting consistent incase of continueExceeding and exponentialBackoff

gurgunday (approved):
lgtm, @Uzlopak might have some other ideas to implement this though

aniketcodes:
@Uzlopak Can you suggest a better approach for this?

Uzlopak:
Will review it today :)

README.md
```diff
@@ -148,7 +148,7 @@ await fastify.register(import('@fastify/rate-limit'), {
...
 - `redis`: by default, this plugin uses an in-memory store, but if an application runs on multiple servers, an external store will be needed. This plugin requires the use of [`ioredis`](https://github.com/redis/ioredis).<br> **Note:** the [default settings](https://github.com/redis/ioredis/blob/v4.16.0/API.md#new_Redis_new) of an ioredis instance are not optimal for rate limiting. We recommend customizing the `connectTimeout` and `maxRetriesPerRequest` parameters as shown in the [`example`](https://github.com/fastify/fastify-rate-limit/tree/master/example/example.js).
 - `nameSpace`: choose which prefix to use in the redis, default is 'fastify-rate-limit-'
-- `continueExceeding`: Renew user limitation when user sends a request to the server when still limited
+- `continueExceeding`: Renew user limitation when user sends a request to the server when still limited. This will take priority over exponentialBackoff
```
Fdawgs, line 151:
```suggestion
- `continueExceeding`: Renew user limitation when user sends a request to the server when still limited. This will take priority over `exponentialBackoff`
```

Fdawgs (changes requested):

aniketcodes:
@Uzlopak @climba03003 @Fdawgs Guys, can you please review?

store/RedisStore.js
```diff
@@ -20,13 +26,20 @@ const lua = `
   if ttl == -1 or (continueExceeding and current > max) then
     redis.call('PEXPIRE', key, timeWindow)
     ttl = timeWindow
+  
+  -- If the key is new or if its incremented value has exceeded the max value and exponential backoff is enabled then set its TTL
+  elseif ttl == -1 or (exponentialBackoff and current > max) then
```
gurgunday, line 31: I'll merge this but you can remove the `ttl == -1` here since the first condition will handle that condition

```suggestion
  elseif exponentialBackoff and current > max then
```

Fdawgs (approved):
Docs wise it lgtm

types/index.d.ts
```diff
@@ -126,6 +126,8 @@ declare namespace fastifyRateLimit {
     enableDraftSpec?: boolean;
     onExceeding?: (req: FastifyRequest, key: string) => void;
     onExceeded?: (req: FastifyRequest, key: string) => void;
+    exponentialBackoff?: boolean;
```
gurgunday, line 129: Please add a type test :)

store/RedisStore.js
```diff
@@ -20,13 +26,20 @@ const lua = `
   if ttl == -1 or (continueExceeding and current > max) then
     redis.call('PEXPIRE', key, timeWindow)
     ttl = timeWindow
+  
+  -- If the key is new or if its incremented value has exceeded the max value and exponential backoff is enabled then set its TTL
+  elseif exponentialBackoff and current > max then
+    local backoffExponent = current - max - 1
+    ttl = math.min(timeWindow * (2 ^ backoffExponent), MAX_SAFE_INTEGER)
+    redis.call('PEXPIRE', key, ttl)
```
gurgunday, line 34: We keep incrementing `current`, so what happens if we surpass the max integer in Lua? It doesn't become Infinity but actually flips to negative

Consider this:

```lua
local max = math.maxinteger
print (max * 2) // -2
```

Then `math.min` will take -2 as `ttl`

The easiest solution is to cast to float, which also makes the number become Infinite once it's out of bounds

gurgunday: This is why we should add a Redis test as well

gurgunday: I recommend doing this at the top:

```lua
local timeWindow = tonumber(ARGV[1]) * 1.0
```

gurgunday (changes requested):

store/RedisStore.js
```diff
@@ -20,13 +26,20 @@ const lua = `
   if ttl == -1 or (continueExceeding and current > max) then
     redis.call('PEXPIRE', key, timeWindow)
     ttl = timeWindow
+  
+  -- If the key is new or if its incremented value has exceeded the max value and exponential backoff is enabled then set its TTL
+  elseif exponentialBackoff and current > max then
+    local backoffExponent = current - max - 1
+    ttl = math.min(timeWindow * (2 ^ backoffExponent), MAX_SAFE_INTEGER)
```
gurgunday, line 33:
```suggestion
    ttl = math.min(1.0 * timeWindow * (2 ^ backoffExponent), MAX_SAFE_INTEGER)
```

store/RedisStore.js
aniketcodes: Done

gurgunday (approved):

gurgunday:
Thanks for the PR and taking the time to address the feedback

PR #394, opened 2024-11-16
fix: `after` should show the same ttl as the rate limit headers

gurgunday:
Simplifies the rate limiter by only looking at the `ttl` to construct the `after` message — this aligns the message with the `Retry-After` header; before, it would just send the `timeWindow`, which is inaccurate

So even if a user had 5 seconds left, if the `timeWindow` was 50 seconds, the error would say, `"Retry in 50 seconds"` and not `"Retry in 5 seconds"`

### DIFF
--- a/index.js
+++ b/index.js
@@ -201,11 +201,6 @@ function addRouteRateHook (pluginComponent, params, routeOptions) {
 function rateLimitRequestHandler (pluginComponent, params) {
   const { rateLimitRan, store } = pluginComponent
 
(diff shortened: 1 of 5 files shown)


gurgunday:
I will rebase after https://github.com/fastify/fastify-rate-limit/pull/393 merges

zekth (approved):
:v:

dancastillo (approved):
LGTM

PR #397, opened 2024-12-08
build(deps-dev): replace standard with neostandard

Fdawgs:
See https://github.com/fastify/fastify-cors/pull/340. This is a batch PR, so may have some issues. Please review changes.


Fdawgs:
Thanks for sorting these @Uzlopak, I'm just about to go on holiday so won't be able to sort them myself until Friday.

Issue #408, opened 2025-01-12
Adding support for query param based rate limit [graphql based servers]


umakantp:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Most GraphQL requests are POST and every request has the same URL, only what changes is a query param value, commonly `operation`. Adding support for rating limiting based on query param instead of the route.

I was not able to locate a proper example of such an example. If it is supported and I could not find out, I guess we can close the issue, if you can share one.

### Motivation

I have been using `mercurius` with `fastify` for the GraphQL server, I want to rate limit some critical paths. `fastify-rate-limit` does pretty good job, if you want route-based rate limiting. I was looking to rate limit based on query param. The server gets `operation` as a query param with values such as `signIn` or `forgetPassword` and so on.

### Example

May not be very useful example signature, but guess send the message across.

```
await app.register(rateLimit, {
  global: true,
  timeWindow: 60 * 1000,
  max: 1000,
  queryBasedRateLimit: {
      [{ operation: ['signIn', 'signUp'], rateLimite: { timeWindow: 6 * 1000....etc. } }],
      [{ operation: ['verifyEmail', 'resendValidationEmail'], rateLimite: { timeWindow: 1 * 1000....etc. } }],
  }  
} )
```

umakantp:
Just in case, I don't know but may be this [PR](https://github.com/fastify/fastify-rate-limit/pull/392), could also help in such a scenario,

mcollina:
@umakantp I think the best approach is https://github.com/fastify/fastify-rate-limit/pull/392. Would you like to take over that PR?

umakantp:
Sure, I can help. I will get it updated first, tested. Thanks

umakantp:
@mcollina Did you get the chance to look at the PR?

mcollina:
Sorry for the long wait, I have been busy :(

umakantp:
Recent release fixes this

PR #392, opened 2024-11-10
feat: add function to manually check rate limit

Charioteer:
Hey fastify contributors,

Following on from issue #346 and my [comment](https://github.com/fastify/fastify-rate-limit/issues/346#issuecomment-2331639190), I finally had some spare time to implement a draft version of my idea as suggested by @mcollina. I realized that any store (child) instance receives all the `options` provided to `fastify.rateLimit` when it is constructed, which is why I had to adjust my initial idea and come up with a slightly different solution.

I did not provide any documentation or additional tests yet (all existing tests run green tho), because I wanted to ask for your feedback first.

Please let me know what you think and as soon as everything looks good to you, I will finalize my PR with proper docs and tests.

Thanks and cheers
Patrick

# How to Use

```ts
import Fastify from "fastify";
import { fastifyRateLimit } from "@fastify/rate-limit";

const fastify = Fastify();

await fastify.register(fastifyRateLimit, {
  global: false,
  max: 10,
  timeWindow: "10 seconds",
});

const checkRateLimit = fastify.createRateLimit(); // this will use the global options provided to fastifyRateLimit

fastify.get("/", async (request, reply) => {
  const limit = await checkRateLimit(request);

  if(!limit.isAllowed && limit.isExceeded) {
    return reply.code(429).send("Limit exceeded");
  }

  return reply.send("Hello world");
});


const checkCustomRateLimit = fastify.createRateLimit({ max: 100 }); // max provided to override global options

fastify.get("/custom", async (request, reply) => {
  const limit = await checkCustomRateLimit(request);

  if(!limit.isAllowed && limit.isExceeded) {
    return reply.code(429).send("Limit exceeded");
  }

  return reply.send("Hello world");
});
```

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

Issue #346, opened 2023-12-12
Add parser option


Kimblis:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

If you use fastify with trpc you most likely use some kind of parser. We are using fastify + trpc + zod as a parser. There is an option for response `errorResponseBuilder` but it expect an object. So we can't fully specify what we return (in our case it should be either TrpcError or stringified json for zod), so that's the only reason we cant currently use this package. (Unless you can think of an easier solution for this, would be very helpful). It would be nice to have a parser option which would apply specified parser to the response (you can include just a couple, the most popular ones)

mcollina:
I don't understand what parser are you referring to. If you would like to add an option so you can plug in some trpc logic, I'm +1, bu I don't think it belongs to this module.

gurgunday:
I don’t get it either, can you please elaborate? Thanks!

Charioteer:
Hi @mcollina and @gurgunday,

I just stumbled accross this issue having a similar one. I am also using `trpc` with `fastify` and `@fastify/rate-limit` to rate limit my tRPC prodcedures (as well as REST API endpoints, more on this later). What I am going to explain in detail now is what I think @Kimblis wanted to ask about initially.

Whenever a [tRPC client](https://trpc.io/docs/client) calls a tRPC procedure "protected" by `@fastify/rate-limit` and, at some point, reaches the configured rate limit, `@fastify/rate-limit` returns a JSON error response as usual. Unfortunately, this leads to the following tRPC error on the client: `TRPCClientError: Unable to transform response from server`. A tRPC client expects errors created with [`TRPCError`](https://trpc.io/docs/server/error-handling). In a best case scenario, one would never return any fastify-related errors in a tRPC procedure (i.e. plain text/JSON reponses) but only `TRPCErrors` to avoid this exact issue. 

Usually, a rate limiter for tRPC would be implemented by using [middlewares](https://trpc.io/docs/server/middlewares). This way, we could throw a `TRPCError` with the proper `TOO_MANY_REQUEST` code. As a side effect, we would also be able to granularly setup rate limiting on a per-procedure level when using middlewares. It would be great if `@fastify/rate-limit` would expose some method to call inside a middleware to check whether a given IP (or any key as returned by `keyGenerator`) already reached its rate limit. This is some rough example for what I would love to be able to do:

```ts
[...]

PR #410, opened 2025-01-19
Add function to manually check rate limit #346 #408

umakantp:
This is the updated PR (as I did not have access to the original PR) copy of [392](https://github.com/fastify/fastify-rate-limit/pull/392). It is a copy with an added test case on work done by @Charioteer and uses the latest master. I did not change any code. Just did some testing and added a unit test case.

The reason for finishing up the PR is issue no [408](https://github.com/fastify/fastify-rate-limit/issues/408). I'm copying the rest of the description here written by  @Charioteer so everything stays in one place. Feel free to close if you can provide me access to the original branch to make the change & wrap it up.

I can add the documentation given if this PR looks good to the contributors.

> From @Charioteer :

Following on from issue #346 and my [comment](https://github.com/fastify/fastify-rate-limit/issues/346#issuecomment-2331639190), I finally had some spare time to implement a draft version of my idea as suggested by @mcollina. I realized that any store (child) instance receives all the `options` provided to `fastify.rateLimit` when it is constructed, which is why I had to adjust my initial idea and come up with a slightly different solution.

Thanks and cheers Patrick

# How to Use
```ts
import Fastify from "fastify";
import { fastifyRateLimit } from "@fastify/rate-limit";

const fastify = Fastify();

await fastify.register(fastifyRateLimit, {
  global: false,
  max: 10,
  timeWindow: "10 seconds",
});

const checkRateLimit = fastify.createRateLimit(); // this will use the global options provided to fastifyRateLimit

fastify.get("/", async (request, reply) ={
  const limit = await checkRateLimit(request);

  if(!limit.isAllowed && limit.isExceeded) {
    return reply.code(429).send("Limit exceeded");
  }

  return reply.send("Hello world");
});


const checkCustomRateLimit = fastify.createRateLimit({ max: 100 }); // max provided to override global options

fastify.get("/custom", async (request, reply) ={
  const limit = await checkCustomRateLimit(request);

  if(!limit.isAllowed && limit.isExceeded) {
    return reply.code(429).send("Limit exceeded");
  }

  return reply.send("Hello world");
});
```

#### Checklist
* [x]  run `npm run test` and `npm run benchmark`
* [x]  tests and/or benchmarks are included
* [ ]  documentation is changed or added
* [x]  commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/test/manual-rate-limit.test.js
+++ b/test/manual-rate-limit.test.js
@@ -25,7 +25,7 @@ describe('Manual check rate limit', () => {
       } else {
         t.assert.deepStrictEqual(limit.isExceeded, false)
       }
-      reply
+      return reply
         .code(limit.isExceeded ? 429 : 200)
         .header('x-ratelimit-limit', limit.max)
         .header('x-ratelimit-remaining', limit.remaining)
--- a/test/manual-rate-limit.test.js
+++ b/test/manual-rate-limit.test.js
@@ -1,0 +1,56 @@
+'use strict'
+
+const { describe, test, mock } = require('node:test')
+const Fastify = require('fastify')
+const { fastifyRateLimit } = require('../index')
+
+describe('Manual check rate limit', () => {
+  test('Basic limit checks', async (t) => {
+    t.plan(13)
+    const clock = mock.timers
+    clock.enable(0)
+    const fastify = Fastify()
+    await fastify.register(fastifyRateLimit, { global: false })
+
+    const checkRateLimit = fastify.createRateLimit({ max: 2, timeWindow: 1000 })
+
+    fastify.get('/', async (req, reply) => {
+      const limit = await checkRateLimit(req)
+      const iteration = req.query.request
+      console.log('req', req.query.request)
+      console.log('limit', limit)
+      t.assert.deepStrictEqual(limit.isAllowed, false)
+      if (iteration === '3') {
+        t.assert.deepStrictEqual(limit.isExceeded, true)
+      } else {
+        t.assert.deepStrictEqual(limit.isExceeded, false)
+      }
+      reply
+        .code(limit.isExceeded ? 429 : 200)
+        .header('x-ratelimit-limit', limit.max)
+        .header('x-ratelimit-remaining', limit.remaining)
+        .header('x-ratelimit-reset', limit.reset)
+        .send('hello!')
+    })
+
+    let res
+
+    res = await fastify.inject('/?request=1')
+
+    t.assert.deepStrictEqual(res.statusCode, 200)
+    t.assert.deepStrictEqual(res.headers['x-ratelimit-limit'], '2')
+    t.assert.deepStrictEqual(res.headers['x-ratelimit-remaining'], '1')
+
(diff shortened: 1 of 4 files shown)


mcollina:
Why is still in draft?

test/manual-rate-limit.test.js
```diff
@@ -0,0 +1,56 @@
...
+      const iteration = req.query.request
+      console.log('req', req.query.request)
+      console.log('limit', limit)
+      t.assert.deepStrictEqual(limit.isAllowed, false)
+      if (iteration === '3') {
+        t.assert.deepStrictEqual(limit.isExceeded, true)
+      } else {
+        t.assert.deepStrictEqual(limit.isExceeded, false)
+      }
+      reply
```
mcollina, line 28:
```suggestion
      return reply
```

types/index.d.ts
```diff
@@ -12,6 +12,23 @@ import {
 
 declare module 'fastify' {
   interface FastifyInstance<RawServer, RawRequest, RawReply, Logger, TypeProvider> {
+    createRateLimit(options?: fastifyRateLimit.CreateRateLimitOptions): (req: FastifyRequest) => Promise<
```
mcollina, line 15: Can you add a test for the types?

mcollina (changes requested):
we use tsd to tests our types.

types/index.d.ts
umakantp: sure

Charioteer:
Hi,

thanks @umakantp for completing/continuing my PR #392. As you may have noticed, I've been busy the whole time and didn't find any spare time yet. I was hoping to complete it in the 1st or 2nd week of February. Regardless of whether you finalize this “new” PR now, I would be very grateful if the authorship of my original code would remain visible in the commit history, which is currently not the case with a merge of #410. But I don't know to what extent this is possible now (@mcollina ?) and in case of doubt it is the way it is, I had enough time to finish it myself and didn't 😄. I also don't think it's a good idea to continue working on #392 now, because it would - vise-versa - collide with your work.

umakantp:
Hi @Charioteer 

I'm fine passing on original commits authorship to you (if possible). You are the owner, after all, and that is why I wrote for history and tagged you in this PR, so there is always your reference for ownership of the code to you. I'm fine even passing you ownership of this PR (if it also transfers the commit's ownership).  All I care for is the feature to be released & hence took over to complete & get it out. So up to you and @mcollina, how you want to proceed with this.

mcollina:
@umakantp why dind't you use the original commits from @Charioteer ?

umakantp:
> @umakantp why dind't you use the original commits from @Charioteer ?

1. I was not able to use the original PR, since I did not had the edit access.
2. If you mean by using the original commits means cherry-picking (or forking his branch to use)? Sorry that did not really strike me at that point.

Looks like this PR is more of problem than a solution for everyone. I'll close it. Will wait for anyone else willing to write a new code or @Charioteer to finish his original PR. Thank you.

PR #413, opened 2025-03-23
feat: node-redis

alesblaznik:
#### Checklist

- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ const { parse, format } = require('@lukeed/ms')
 
 const LocalStore = require('./store/LocalStore')
 const RedisStore = require('./store/RedisStore')
+const NodeRedisStore = require('./store/NodeRedisStore')
 
 const defaultMax = 1000
 const defaultTimeWindow = 60000
@@ -117,7 +118,11 @@ async function fastifyRateLimit (fastify, settings) {
     pluginComponent.store = new Store(globalParams)
   } else {
     if (settings.redis) {
-      pluginComponent.store = new RedisStore(globalParams.continueExceeding, globalParams.exponentialBackoff, settings.redis, settings.nameSpace)
+      if (settings.redis.constructor.name === 'Commander') {
+        pluginComponent.store = new NodeRedisStore(globalParams.continueExceeding, globalParams.exponentialBackoff, settings.redis, settings.nameSpace)
+      } else {
+        pluginComponent.store = new RedisStore(globalParams.continueExceeding, globalParams.exponentialBackoff, settings.redis, settings.nameSpace)
+      }
     } else {
       pluginComponent.store = new LocalStore(globalParams.continueExceeding, globalParams.exponentialBackoff, settings.cache)
     }
@@ -290,3 +295,4 @@ module.exports = fp(fastifyRateLimit, {
 })
 module.exports.default = fastifyRateLimit
 module.exports.fastifyRateLimit = fastifyRateLimit
+module.exports.NodeRedisStore = NodeRedisStore
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "lint:fix": "eslint --fix",
     "redis": "docker run -p 6379:6379 --name rate-limit-redis -d --rm redis",
     "test": "npm run test:unit && npm run test:typescript",
-    "test:unit": "c8 --100 node --test",
+    "test:unit": "c8 --100 node --test --test-concurrency=1",
     "test:typescript": "tsd"
   },
   "repository": {
@@ -60,6 +60,7 @@
   ],
   "devDependencies": {
     "@fastify/pre-commit": "^2.1.0",
+    "@redis/client": "^1.6.0",
     "@sinonjs/fake-timers": "^14.0.0",
     "@types/node": "^22.0.0",
     "c8": "^10.1.2",
--- a/store/NodeRedisStore.js
+++ b/store/NodeRedisStore.js
@@ -1,0 +1,52 @@
+'use strict'
+
+/**
+ * When using node-redis, you need to initialize the client with the rateLimit script like this:
+ * ```js
+ * const redis = createClient({
+ *   scripts: {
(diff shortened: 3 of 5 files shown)


gurgunday:
This already looks pretty great :)

index.js
```diff
@@ -117,7 +118,11 @@ async function fastifyRateLimit (fastify, settings) {
     pluginComponent.store = new Store(globalParams)
   } else {
     if (settings.redis) {
-      pluginComponent.store = new RedisStore(globalParams.continueExceeding, globalParams.exponentialBackoff, settings.redis, settings.nameSpace)
+      if (settings.redis.constructor.name === 'Commander') {
```
alesblaznik, line 121: Is there any better way to detect `node-redis`? 

As far as I checked, but need to confirm, all clients (including Redis sentinel) extend from Commander.

- [ ] Double check this.

store/NodeRedisStore.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+const { lua } = require('./RedisStore')
+const { defineScript } = require('@redis/client')
+
+const rateLimitScript = defineScript({
+  NUMBER_OF_KEYS: 1,
+  SCRIPT: lua,
+  transformArguments (key, timeWindow, max, continueExceeding, exponentialBackoff) {
+    return [key, timeWindow.toString(), max.toString(), continueExceeding.toString(), exponentialBackoff.toString()]
```
gurgunday, line 21: Since multiple types can be passed here, I'd recommend using the `String` constructor function or backticks here

When you use a method V8 tries to inline it but if the type changes it triggers a deopt

package.json
```diff
@@ -10,7 +10,7 @@
     "lint:fix": "eslint --fix",
     "redis": "docker run -p 6379:6379 --name rate-limit-redis -d --rm redis",
     "test": "npm run test:unit && npm run test:typescript",
-    "test:unit": "c8 --100 node --test",
+    "test:unit": "c8 --100 node --test --test-concurrency=1",
```
alesblaznik, line 13: Use other redis db instead!

store/RedisStore.js
```diff
@@ -62,3 +62,4 @@ RedisStore.prototype.child = function (routeOptions) {
 }
 
 module.exports = RedisStore
+module.exports.lua = lua
```
alesblaznik, line 65: Need `lua` script for `NodeRedisStore` instead of copy&paste. Both of those should always be the same for all Redis clients.

Should we move lua script to other place instead?

test/node-redis-rate-limit.test.js
alesblaznik, line 1: Copy pasted tests for `ioredis`... should rather just reuse tests and run them with other client instead?

alesblaznik:
> This already looks pretty great :)

Will make comments as there are many things we might want to change... just started with the initial PR and can then discuss how we want to do it.

index.js
gurgunday: Rate limiter flexible, a very popular lib, does it similarly

https://github.com/animir/node-rate-limiter-flexible/blob/0840cd7ebee83e9b17b8f27b53b77583dc143076/lib/RateLimiterRedis.js#L32

alesblaznik: Thanks 🙏🏻

store/NodeRedisStore.js
gurgunday:
```suggestion
    return [key, String(timeWindow), String(max), String(continueExceeding), String(exponentialBackoff)]
```

store/NodeRedisStore.js
```diff
@@ -0,0 +1,52 @@
...
+  transformReply (reply) {
+    return reply
+  },
+})
+
+function NodeRedisStore (continueExceeding, exponentialBackoff, redis, key = 'fastify-rate-limit-') {
+  this.continueExceeding = continueExceeding
+  this.exponentialBackoff = exponentialBackoff
+  this.redis = redis
+  this.key = key
```
alesblaznik, line 32: Should throw an Error here with hint how to `createClient` - passing the script?

store/NodeRedisStore.js
```diff
@@ -0,0 +1,59 @@
...
+
+  if (!this.redis.rateLimit) {
+    throw new Error(
+      'rateLimit script missing on Redis instance. Add it when creating client: ' +
+      'const redis = createClient({ scripts: { rateLimit: rateLimit.NodeRedisStore.rateLimitScript }})'
+    )
+  }
+}
```
gurgunday, line 40: My only gripe with this is that we now expose the name of the script while with ioredis we'd load it internally. While the idea of injecting your own script seems cool, it currently seems a little inconsistent to me

Can't we initialize or use the script after the client is created?

`.eval` seems like an option:

https://github.com/animir/node-rate-limiter-flexible/blob/0840cd7ebee83e9b17b8f27b53b77583dc143076/lib/RateLimiterRedis.js#L121-L125

store/NodeRedisStore.js
```diff
@@ -0,0 +1,59 @@
...
+ * const redis = createClient({
+ *   scripts: {
+ *     rateLimit: rateLimit.NodeRedisStore.rateLimitScript
+ *   }
+ * });
+ * ```
+ */
+
+const { lua } = require('./RedisStore')
+const { defineScript } = require('@redis/client')
```
mcollina, line 15: I would not add a depedency to `@redis/client` here. we should be taking an instance of that in the constructor instead.

gurgunday: Yeah same as https://github.com/fastify/fastify-rate-limit/pull/413#discussion_r2009189455

The goal is to not depend on the lib

Only the client that's passed should be used

PR #437, opened 2026-03-16
feat: Add first-class Valkey support with GLIDE-backed store

janisto:
## Why

Applications using Valkey through GLIDE should be able to share their existing client with `@fastify/rate-limit`, instead of maintaining a separate Redis connection or implementing a custom store.

Closes #436.

Related: https://github.com/fastify/fastify-valkey-glide/pull/33

## What changed

- Add a `valkey` option accepting either a `GlideClient` or `GlideClusterClient`.
- Add a Valkey-backed store using the same Lua behavior as the existing Redis store.
- Use GLIDE's documented [`Script` and `invokeScript` APIs](https://glide.valkey.io/migration/nodejs/ioredis/command-comparison-ioredis-glide/#custom-commands-1). GLIDE therefore handles script loading, caching, and cache-miss recovery; no raw `customCommand` fallback is maintained.
- Create the two fixed scripts lazily and retain them for the module lifetime.
- Keep `store` precedence unchanged and reject simultaneous `redis` and `valkey` options when no custom store is provided.
- Add structural TypeScript support without making GLIDE mandatory for users who do not use Valkey.
- Use `@valkey/valkey-glide` `^2.0.1`, matching `@fastify/valkey-glide`.
- Add focused store, plugin, type, standalone integration, and contributor documentation.

## GLIDE installation

GLIDE associates `Script` instances with its module-level script registry. The supplied client and the scripts therefore need to resolve from the same application-level `@valkey/valkey-glide` installation.

The README installs `@valkey/valkey-glide` at the application root alongside `@fastify/valkey-glide`. This is the supported configuration. The implementation intentionally does not add recovery machinery for applications loading multiple physical GLIDE copies.

## Valkey CI follow-up

Fastify currently provides a reusable Redis workflow through [`plugins-ci-redis.yml`](https://github.com/fastify/workflows/blob/main/.github/workflows/plugins-ci-redis.yml), but no equivalent Valkey workflow.

The real Valkey integration test in this PR is therefore skipped unless `VALKEY_HOST` is set. It does not use `REDIS_HOST` or run against the Redis service: that would validate Redis compatibility rather than Valkey, and the Redis 5 lane is below [GLIDE's supported Redis versions](https://github.com/valkey-io/valkey-glide?tab=readme-ov-file#supported-engine-versions).

Recommended follow-up:

1. Add `plugins-ci-valkey.yml` to `fastify/workflows`, based on the existing reusable Redis workflow but provisioning actual Valkey.
2. Expose the service through `VALKEY_HOST`.
3. Consume that workflow from this repository to enable the real integration test.
4. Reuse the same workflow in `@fastify/valkey-glide`.

Until that workflow exists, the Valkey store and plugin tests continue to run in normal CI, while the real service test remains visibly skipped.

## Validation

- `VALKEY_HOST=127.0.0.1 npm test`
- Real standalone Valkey 9.1.1 with GLIDE 2.5.1
- Standalone smoke test with the minimum supported GLIDE 2.0.1
- Real three-primary Valkey cluster smoke test with `GlideClusterClient`


janisto:
I’ve rebased this PR onto the current main and substantially reworked the implementation based on the earlier feedback.

The updated approach uses GLIDE’s Script and invokeScript APIs, keeps GLIDE optional for users who do not use Valkey, and runs the real Valkey integration test only when VALKEY_HOST is available. The PR description also documents the shared GLIDE installation requirement and proposes a reusable Fastify Valkey CI workflow instead of testing Valkey behavior against Redis.

I’d appreciate feedback on the public valkey option, the script lifecycle approach, and the proposed CI direction when you have time.

PR #443, opened 2026-04-25
refactor(types): migrate from tsd to tstyche

Tony133:
Proposal:

- migrate from `tsd` to `tstyche`

PR of Reference: 
  - https://github.com/fastify/fastify/pull/6532
  - https://github.com/fastify/fastify/pull/6525

### DIFF
--- a/types/index.tst.ts
+++ b/types/index.tst.ts
@@ -60,13 +60,13 @@ const options1: RateLimitPluginOptions = {
       return {
         statusCode: 403,
         error: 'Forbidden',
-        message: `You can not access this service as you have sent too many requests that exceed your rate limit. Your IP: ${req.ip} and Limit: ${context.max}`
+        message: `You can not access this service as you have sent too many requests that exceed your rate limit. Your IP: ${req.ip} and Limit: ${context.max}`,
       }
     } else {
       return {
         statusCode: 429,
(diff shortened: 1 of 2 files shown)


types/index.tst.ts
```diff
@@ -192,24 +234,28 @@ appWithHttp2.register(fastifyRateLimit, options7)
...
 
-expectAssignable<errorResponseBuilderContext>({
+expect({
   statusCode: 429,
   ban: true,
   after: '123',
   max: 1000,
   ttl: 123
-})
+}).type.toBeAssignableTo<errorResponseBuilderContext>()
```
mrazauskas: Use `.toBeAssignableFrom()` instead, to keep the type under tests on the left-hand side of the assertion:

```suggestion
expect<errorResponseBuilderContext>().type.toBeAssignableFrom({
  statusCode: 429,
  ban: true,
  after: '123',
  max: 1000,
  ttl: 123
})
```

Tony133: see commit here: [c98110a](https://github.com/fastify/fastify-rate-limit/pull/443/commits/c98110ae638f99f40f36d42be7da9ec722b3c510)

mrazauskas:
lgtm

PR #446, opened 2026-05-31
Fix Unsafe SQL Queries in Knex Example Files

MatrixNeoKozak:
The Knex example files (`example/example-knex-mysql.js` and `example/example-knex.js`) contained unsafe raw SQL queries that directly interpolated user-supplied values (`this.route` and `key`) into SQL strings, making them vulnerable to SQL injection. This change replaces those raw queries with proper parameterized queries using Knex's binding syntax, which prevents injection attacks and improves safety. The changes are necessary because the examples are distributed as part of the package and could be used as reference by users.

### DIFF
--- a/example/example-knex-mysql.js
+++ b/example/example-knex-mysql.js
@@ -54,19 +54,26 @@ KnexStore.prototype.incr = async function (key, cb) {
   try {
     // NOTE: MySQL syntax FOR UPDATE for read lock on counter stats in row
     const row = await trx('rate_limits')
-      .whereRaw('route = ? AND source = ? FOR UPDATE', [cond.route || '', cond.source]) // Create read lock
+      .where('route', cond.route || '')
+      .andWhere('source', cond.source)
+      .forUpdate()
+      .select()
     const d = row[0]
     if (d && d.ttl > now) {
       // Optimization - no need to UPDATE if max has been reached.
       if (d.count < max) {
-        await trx
-          .raw('UPDATE rate_limits SET count = ? WHERE route = ? AND source = ?', [d.count + 1, cond.route, key])
+        await trx('rate_limits')
+          .where('route', cond.route)
+          .andWhere('source', key)
+          .update({ count: d.count + 1 })
       }
       // If we were already at max no need to UPDATE but we must still send d.count + 1 to trigger rate limit.
       process.nextTick(cb, null, { current: d.count + 1, ttl: d.ttl })
     } else {
-      await trx
-        .raw('INSERT INTO rate_limits(route, source, count, ttl) VALUES(?,?,1,?) ON DUPLICATE KEY UPDATE count = 1, ttl = ?', [cond.route, key, d?.ttl || ttl, ttl])
+      await trx.raw(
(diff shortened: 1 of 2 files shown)


example/example-knex-mysql.js
```diff
@@ -54,19 +54,26 @@ KnexStore.prototype.incr = async function (key, cb) {
   try {
     // NOTE: MySQL syntax FOR UPDATE for read lock on counter stats in row
     const row = await trx('rate_limits')
-      .whereRaw('route = ? AND source = ? FOR UPDATE', [cond.route || '', cond.source]) // Create read lock
```
jean-michelet, line 57: SQL-injection protected here because this uses parameter bindings.

example/example-knex-mysql.js
```diff
@@ -54,19 +54,26 @@ KnexStore.prototype.incr = async function (key, cb) {
...
+      .where('route', cond.route || '')
+      .andWhere('source', cond.source)
+      .forUpdate()
+      .select()
     const d = row[0]
     if (d && d.ttl > now) {
       // Optimization - no need to UPDATE if max has been reached.
       if (d.count < max) {
-        await trx
-          .raw('UPDATE rate_limits SET count = ? WHERE route = ? AND source = ?', [d.count + 1, cond.route, key])
```
jean-michelet, line 63: SQL-injection protected here because this uses parameter bindings.

example/example-knex-mysql.js
```diff
@@ -54,19 +54,26 @@ KnexStore.prototype.incr = async function (key, cb) {
...
+        await trx('rate_limits')
+          .where('route', cond.route)
+          .andWhere('source', key)
+          .update({ count: d.count + 1 })
       }
       // If we were already at max no need to UPDATE but we must still send d.count + 1 to trigger rate limit.
       process.nextTick(cb, null, { current: d.count + 1, ttl: d.ttl })
     } else {
-      await trx
-        .raw('INSERT INTO rate_limits(route, source, count, ttl) VALUES(?,?,1,?) ON DUPLICATE KEY UPDATE count = 1, ttl = ?', [cond.route, key, d?.ttl || ttl, ttl])
```
jean-michelet, line 69: SQL-injection protected here because this uses parameter bindings.

example/example-knex.js
```diff
@@ -55,7 +55,7 @@ KnexStore.prototype.incr = function (key, cb) {
             })
         } else {
           trx
-            .raw(`INSERT INTO RateLimits(Route, Source, Count, TTL) VALUES('${this.route}', '${key}',1,${d.TTL || ttl}) ON CONFLICT(Route, Source) DO UPDATE SET Count=Count+1,TTL=${ttl}`)
```
jean-michelet, line 58: I agree this query is generally unsafe because it interpolates values directly into the SQL string.
That said, the values appear to come from plugin/user configuration rather than directly from a network client. So much less critical than the PR description suggests.

jean-michelet: Ok for changing this, let's keep the other queries as they are

MatrixNeoKozak:
I have reverted the changes to `example/example-knex-mysql.js` as they were already using safe parameter bindings, and kept the security fixes for the unsafe string interpolations in `example/example-knex.js` as requested.

jean-michelet (approved):

PR #449, opened 2026-06-05
feat: add increment option to createRateLimit

### LINKED ISSUE #420: Add a way to check the status of rate limiters without incrementing them
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

After creating a rate limiter manually using fastify.createRateLimit(), I can call this limiter as a function. However, this also increments the limiter. I would like a way to just check the current status of the rate limiter without incrementing it.

Here's an example syntax of how it might work:

```
// Create manual rate limiter
const checkRateLimit = fastify.createRateLimit();

// Check status and increment limiter, like it works today
const limit = await checkRateLimit(request);

// Proposal to only check status without increment. Returns the same as above, but doesn't change the "remaining" property.
const limit = await checkRateLimit(request, { increment: false });
```

The reason I need this is because I want to apply a rate limiter to my login endpoint specifically, but only want it incremented when the login attempt fails. I still need to check if my limiter has been reached regardless of success/failure, so that I can deny access if it's reached, which is why I need this feature.

Another solution would be a way to reset/modify the "remaining" property manually, so I could reset it when the user successfully logins. Whichever solution is fine by me, but I figured that the ability to check status of a limiter without incrementing it might also be useful for other purposes, such as monitoring.

### Motivation

_No response_

### Example

_No response_

Betouss:
feat: add increment option to createRateLimit

Allow consumers of fastify.createRateLimit() to check the rate limit status without consuming a request, by passing { increment: false } as an optional second argument to the limiter function.

Closes #420

#### Checklist

- [ ] run `npm run test && npm run benchmark --if-present`
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ If `isAllowed` is `false` the object also contains these additional properties:
 - `isExceeded`: `true` if the limit was exceeded.
 - `isBanned`: `true` if the request was banned according to the `ban` option.
 
-The limiter function accepts an optional second argument `{ increment: boolean }`. When `increment` is `false`, the current rate limit status is returned **without consuming a request**. This is useful when a limit should only be enforced on certain outcomes (e.g. failed login attempts) while still checking the status before processing.
+The limiter function accepts an optional second argument `{ increment?: boolean }`. The `increment` flag defaults to `true`, so omitting the argument keeps the original behavior (the request is consumed). When `increment` is `false`, the current rate limit status is returned **without consuming a request**. This is useful when a limit should only be enforced on certain outcomes (e.g. failed login attempts) while still checking the status before processing.
 
 ```js
 const checkRateLimit = fastify.createRateLimit({ max: 5, timeWindow: '1 minute' });
@@ -539,6 +539,12 @@ fastify.post('/login', async (request, reply) => {
   return { ok: true };
 });
 ```
+
+A few things to keep in mind when using `{ increment: false }`:
+
+- **It is a non-mutating snapshot.** A peek never increments the counter, resets the window, or advances the `ban`/`continueExceeding`/`exponentialBackoff` side effects that a real request triggers. The returned `isExceeded`/`isBanned` therefore reflect the *current* counters, but a peek will not, by itself, escalate a ban or extend a backoff window.
+- **`ttl` reflects the store's current window.** With the Redis store, `ttl` is the raw server `PTTL` (the same value `incr` reports), so it can exceed the configured `timeWindow` when `continueExceeding`/`exponentialBackoff` has extended it.
+- **Custom stores must implement `read`.** The flag relies on a non-mutating `read(ip, cb, timeWindow, max)` method, which mirrors the `incr` signature. The built-in local and Redis stores provide it; a custom store that does not will throw a clear error when called with `{ increment: false }`.
 
 ### Examples of Custom Store
 
--- a/README.md
+++ b/README.md
@@ -517,6 +517,29 @@ If `isAllowed` is `false` the object also contains these additional properties:
 - `isExceeded`: `true` if the limit was exceeded.
 - `isBanned`: `true` if the request was banned according to the `ban` option.
 
+The limiter function accepts an optional second argument `{ increment: boolean }`. When `increment` is `false`, the current rate limit status is returned **without consuming a request**. This is useful when a limit should only be enforced on certain outcomes (e.g. failed login attempts) while still checking the status before processing.
+
+```js
+const checkRateLimit = fastify.createRateLimit({ max: 5, timeWindow: '1 minute' });
+
+fastify.post('/login', async (request, reply) => {
+  // Peek at the current status without consuming a request
+  const status = await checkRateLimit(request, { increment: false });
+  if (status.isExceeded) {
+    return reply.code(429).send({ error: 'Too many attempts' });
+  }
+
+  const success = await tryLogin(request.body);
+  if (!success) {
+    // Only consume a request when the login fails
+    await checkRateLimit(request);
+    return reply.code(401).send({ error: 'Invalid credentials' });
+  }
+
+  return { ok: true };
+});
+```
+
 ### Examples of Custom Store
 
 These examples show an overview of the `store` feature and you should take inspiration from it and tweak as you need:
--- a/index.js
+++ b/index.js
@@ -128,7 +128,7 @@ async function fastifyRateLimit (fastify, settings) {
   if (!fastify.hasDecorator('createRateLimit')) {
     fastify.decorate('createRateLimit', (options) => {
       const args = createLimiterArgs(pluginComponent, globalParams, options)
-      return (req) => applyRateLimit.apply(this, args.concat(req))
+      return (req, callOptions) => applyRateLimit.apply(this, args.concat(req, callOptions))
     })
   }
 
@@ -210,7 +210,7 @@ function addRouteRateHook (pluginComponent, params, routeOptions) {
   }
 }
 
-async function applyRateLimit (pluginComponent, params, req) {
+async function applyRateLimit (pluginComponent, params, req, callOptions) {
   const { store } = pluginComponent
 
   // Retrieve the key from the generator (the global one or the one defined in the endpoint)
@@ -244,10 +244,10 @@ async function applyRateLimit (pluginComponent, params, req) {
   let ttl = 0
   let ttlInSeconds = 0
 
-  // We increment the rate limit for the current request
+  const storeMethod = callOptions?.increment === false ? 'read' : 'incr'
   try {
     const res = await new Promise((resolve, reject) => {
-      store.incr(key, (err, res) => {
+      store[storeMethod](key, (err, res) => {
         err ? reject(err) : resolve(res)
       }, timeWindow, max)
     })
--- a/index.js
+++ b/index.js
@@ -245,6 +245,14 @@ async function applyRateLimit (pluginComponent, params, req, callOptions) {
   let ttlInSeconds = 0
 
   const storeMethod = callOptions?.increment === false ? 'read' : 'incr'
+
+  // `{ increment: false }` requires the store to implement a non-mutating
+  // `read` method. Custom stores may not, so fail fast with a clear message
+  // instead of a cryptic "store[storeMethod] is not a function" TypeError.
+  if (storeMethod === 'read' && typeof store.read !== 'function') {
+    throw new Error('The configured rate-limit store does not implement a `read` method, which is required to use `{ increment: false }`')
+  }
+
   try {
     const res = await new Promise((resolve, reject) => {
       store[storeMethod](key, (err, res) => {
--- a/store/LocalStore.js
+++ b/store/LocalStore.js
@@ -43,7 +43,25 @@ LocalStore.prototype.incr = function (ip, cb, timeWindow, max) {
   cb(null, current)
 }
 
-LocalStore.prototype.read = function (ip, cb, timeWindow) {
+/**
+ * Read the current rate-limit state for `ip` without mutating it.
+ *
+ * Stores expose `read` with the same argument contract as `incr`
+ * (`ip, cb, timeWindow, max`) so the two are interchangeable; an
+ * implementation may ignore the arguments it does not need (`max` here).
+ *
+ * `read` is a non-mutating snapshot: it never increments the counter, resets
+ * the window, or advances the `continueExceeding`/`exponentialBackoff`/`ban`
+ * side effects that `incr` applies. It mirrors `incr`'s window-expiry
+ * detection, so a peek and a real request agree on whether the window is
+ * still active.
+ *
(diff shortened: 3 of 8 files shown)


Betouss:
Implements #420 as discussed. @jean-michelet happy to adjust anything.

jean-michelet:
Thx, I am not familiar with the source code of this plugin so I prefer your work to be reviewed by someone else from @fastify/plugins

store/LocalStore.js
```diff
@@ -43,6 +43,21 @@ LocalStore.prototype.incr = function (ip, cb, timeWindow, max) {
   cb(null, current)
 }
 
+LocalStore.prototype.read = function (ip, cb, timeWindow) {
```
ilteoood: Why the signature of the read function is different between the local store and the redis store?

ilteoood (commented):
I have a single question at the moment. What comes next is an AI review. It could make mistakes.

store/LocalStore.js
```diff
@@ -43,6 +43,21 @@ LocalStore.prototype.incr = function (ip, cb, timeWindow, max) {
...
+  const nowInMs = Date.now()
+  const current = this.lru.get(ip)
+
```
ilteoood, line 67: Signature inconsistency: this takes `(ip, cb, timeWindow)` while `RedisStore.prototype.read` takes `(ip, cb)`. The `incr` contract is uniform across both stores (`(ip, cb, timeWindow, max)`), so the asymmetry on `read` means any third-party store author has to guess. Either align both signatures and ignore unused args, or document the contract (ideally as JSDoc on the store base) so custom stores can implement `read` correctly.

store/RedisStore.js
```diff
@@ -51,6 +80,12 @@ RedisStore.prototype.incr = function (ip, cb, timeWindow, max) {
   })
 }
 
+RedisStore.prototype.read = function (ip, cb) {
+  this.redis.rateLimitRead(this.key + ip, (err, result) => {
+    err ? cb(err, null) : cb(null, { current: result[0], ttl: result[1] })
```
ilteoood, line 101: Same as the comment on `LocalStore.prototype.read`: this omits `timeWindow` from the signature, breaking the symmetry with `incr` and with the `LocalStore` counterpart. If the intent is "we don't need `timeWindow` here because PTTL is the source of truth", that's fine — but the contract for custom store authors should be explicit (uniform signature, with unused args allowed and ignored).

index.js
```diff
@@ -244,10 +244,10 @@ async function applyRateLimit (pluginComponent, params, req) {
   let ttl = 0
   let ttlInSeconds = 0
 
-  // We increment the rate limit for the current request
+  const storeMethod = callOptions?.increment === false ? 'read' : 'incr'
```
ilteoood, line 247: On the `read` path the counter is read-only, but the return block further down still computes `isBanned: params.ban !== -1 && current - max > params.ban` (line ~273). Since `read` doesn't trigger the ban machinery, a banned IP peeked with `{ increment: false }` will report `isAllowed: false` with the current counter — not `isBanned: true`. This is arguably a deliberate "peek doesn't apply policy" choice, but it's a footgun: a caller peeking to decide whether to consume may incorrectly treat a banned IP as healthy. Either reflect ban state on `read` (read from a separate ban store / counter) or document the limitation explicitly in the README.

store/RedisStore.js
```diff
@@ -31,6 +31,28 @@ const lua = `
...
+  -- Read the counter without mutating it
+  local current = redis.call('GET', key)
+
+  if current == false then
+    -- Key doesn't exist: clean state
+    return {0, 0}
+  end
+
+  -- Read the remaining TTL in milliseconds
+  local ttl = redis.call('PTTL', key)
```
ilteoood, line 49: This script reports whatever PTTL says, without clamping to the configured `timeWindow`. Concretely: if the key exists with a TTL larger than the configured `timeWindow` (e.g. a previously larger window, or a different `timeWindow` configured by an earlier process), `read` will return a `ttl` bigger than the configured `timeWindow` — the caller will think the window is still alive when `incr` would have reset it. `incr` resets the window via PEXPIRE, so the two paths can disagree on the same key. Either clamp `ttl` to `timeWindow` here, or document that `read` reports the raw server state and may diverge from the configured window.

test/create-rate-limit.test.js
```diff
@@ -221,4 +221,104 @@ test('With allow list', async t => {
     isAllowed: true,
     key: '127.0.0.1'
   })
+
+  clock.reset()
+})
+
+test('With { increment: false } it reads the state without consuming it', async t => {
```
ilteoood, line 228: The new test covers the default config but doesn't exercise `continueExceeding: true` (or `exponentialBackoff: true`). With `incr`, both options mutate the TTL when the limit is exceeded (`LocalStore.incr` resets `iterationStartMs` and `ttl`; the Redis `incr` Lua calls `PEXPIRE` again). With `read`, neither mutation happens — which is the whole point — but that also means a peek after exceeding will report the *original* window's remaining TTL, not a reset one. Worth pinning down behaviorally with a dedicated test, since the asymmetry between `incr` and `read` on the TTL is the kind of thing that bites you in production.

index.js
```diff
@@ -244,10 +244,10 @@ async function applyRateLimit (pluginComponent, params, req) {
...
   try {
     const res = await new Promise((resolve, reject) => {
-      store.incr(key, (err, res) => {
+      store[storeMethod](key, (err, res) => {
```
ilteoood, line 258: Once `storeMethod === 'read'`, the function still computes `isBanned`, `isExceeded`, etc. from `current` on the return path — fine, those are still meaningful for the caller. But the `incr`-side invariants those flags rely on (ban accumulation, `continueExceeding` TTL reset, `exponentialBackoff` window expansion) are never applied on `read`, so the `isBanned`/`isExceeded` semantics for a peek are subtly different from a real request. Not a bug, but a small short-circuit opportunity: when `storeMethod === 'read'`, you can skip the `current - max > params.ban` check (ban can't have been triggered by a read) and the README/return shape should make this explicit. Also worth considering: avoid passing `timeWindow` and `max` to `read` at all on the read path, since they're unused by the current implementations.

store/RedisStore.js
```diff
@@ -31,6 +31,28 @@ const lua = `
...
 `
 
+const luaRead = `
+  -- Key to operate on
+  local key = KEYS[1]
+
+  -- Read the counter without mutating it
+  local current = redis.call('GET', key)
+
+  if current == false then
```
gurgunday:
```suggestion
  if current == nil then
```

store/LocalStore.js
Betouss: Same point, both `read` methods now share the `incr` signature `(ip, cb, timeWindow, max)`, with JSDoc documenting the contract. Fixed in 9dc865c.

store/LocalStore.js
Betouss: Good catch. I aligned both stores to `read(ip, cb, timeWindow, max)`, matching the `incr` contract, and added JSDoc on both `read` methods documenting it: same signature as `incr`, non-mutating, and an implementation may ignore the args it doesn't need. Fixed in 9dc865c.

store/RedisStore.js
Betouss: Aligned with the above: `RedisStore.prototype.read` is now `read(ip, cb, timeWindow, max)` too (it ignores `timeWindow`/`max` since `PTTL` is the source of truth), with JSDoc noting it. 9dc865c.

index.js
Betouss: Agreed this deserves to be explicit. I treated `{ increment: false }` as a deliberate non-mutating *snapshot*: it reports current counters (so `isExceeded`/`isBanned` reflect existing state) but never triggers or advances ban/backoff. Reflecting live ban escalation would need a separate ban store the plugin doesn't keep, so I documented the snapshot semantics in the README rather than half-implementing it. Happy to change the behavior if you'd prefer. 9dc865c.

store/RedisStore.js
Betouss: I looked into this: the existing `incr` Lua already returns the raw `PTTL` on its alive path and `PEXPIRE`s beyond `timeWindow` under `continueExceeding`/`exponentialBackoff`. So `read` returning raw `PTTL` is actually *consistent* with `incr`, clamping to `timeWindow` would make `read` diverge from `incr` under backoff. I documented that `read` reports the raw server TTL. Glad to clamp instead if you'd rather. 9dc865c.

test/create-rate-limit.test.js
Betouss: Added dedicated tests for both `continueExceeding: true` and `exponentialBackoff: true` with `{ increment: false }`. They pin the behavior: a peek inside the active window reports the exceeded count and the base-window `ttl` (without escalating backoff), and reports a clean state once the window elapses: matching `incr`'s expiry detection. 9dc865c.

index.js
Betouss: The signature is now uniform with `incr`, and each store ignores what it doesn't need. I kept `isBanned`/`isExceeded` in the return shape (still meaningful as a current-state snapshot) and documented that `read` doesn't advance ban/backoff. I left the ban computation in rather than short-circuiting, to keep the return shape identical between a peek and a real request, but happy to short-circuit if you prefer. 9dc865c.

store/RedisStore.js
Betouss: Thanks! `redis.call('GET')` returns `false` (not `nil`) for a missing key, so `== nil` wouldn't match it. I went with `if not current then`, which is robust to both `false` (`redis.call`) and `nil` (`redis.pcall`) and reads cleanly. Let me know if you'd rather have an explicit `== false`. 9dc865c.

store/LocalStore.js
```diff
@@ -43,6 +43,21 @@ LocalStore.prototype.incr = function (ip, cb, timeWindow, max) {
...
+  if (!current || current.iterationStartMs + timeWindow <= nowInMs) {
+    // Item doesn't exist or has expired: report a clean state without mutating
+    cb(null, { current: 0, ttl: 0 })
+    return
+  }
+
+  // Item is alive: report the current state without mutating
+  const ttl = timeWindow - (nowInMs - current.iterationStartMs)
+  cb(null, { current: current.current, ttl })
+}
```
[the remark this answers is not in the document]
Betouss: `read` uses the same window-expiry check as `incr` (`iterationStartMs + timeWindow`), so it doesn't expire earlier than `incr` or let clients bypass the window. I added tests under `exponentialBackoff` showing a peek still reports the exceeded count while the window is active, and only reports a clean state once `incr` would also reset. The reported `ttl` is the base window by design (a peek doesn't escalate backoff), now documented. 9dc865c.

index.js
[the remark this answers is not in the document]
Betouss: Good call, added a fast-fail with a clear error message when `{ increment: false }` is used with a store that doesn't implement `read`, plus a test and a README note. 9dc865c.

README.md
```diff
@@ -517,6 +517,29 @@ If `isAllowed` is `false` the object also contains these additional properties:
 - `isExceeded`: `true` if the limit was exceeded.
 - `isBanned`: `true` if the request was banned according to the `ban` option.
 
+The limiter function accepts an optional second argument `{ increment: boolean }`. When `increment` is `false`, the current rate limit status is returned **without consuming a request**. This is useful when a limit should only be enforced on certain outcomes (e.g. failed login attempts) while still checking the status before processing.
```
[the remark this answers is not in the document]
Betouss: Updated: the docs now show `{ increment?: boolean }`, clarify it defaults to incrementing, and note that custom stores must implement `read`. 9dc865c.

Betouss:
Thanks for the detailed review, @ilteoood @gurgunday! Addressed everything in 9dc865c — replies inline on each thread. Two points are judgment calls (snapshot semantics for ban/backoff, and raw vs clamped Redis TTL); I explained my reasoning inline and I'm happy to adjust either if you'd prefer.

All checks green locally: lint, `npm run test:unit` (100% coverage, incl. Redis), and `tstyche` types.

ilteoood (approved):

Betouss:
Gentle ping. This has been approved by @ilteoood with green CI across Node 20, 22 and 24, and the branch is mergeable and up to date. Is there anything else needed before it can be merged?

@gurgunday, did my reply about `redis.call('GET')` returning `false` (rather than `nil`) address your suggestion? Happy to make any further changes if needed.

gurgunday:
I will review this again today

gurgunday (approved):
lgtm

Eomm (approved):

PR #380, opened 2024-09-02
Added grouping routes for ratelimit

aniketcodes:
This PR adds the capability to group various routes and have a common rate limit for them.
#369

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #464, opened 2026-09-17
fix: groupId should share rate-limit counters across routes (#463)

### LINKED ISSUE #463: groupId does not group routes: each route in a group gets the full max
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.11.2

### Plugin version

11.2.0

### Node.js version

24.20.0

### Operating system

Linux

Ubuntu 26.04.1 LTS (kernel 7.0.0-30-generic, x86_64)

### Description

Summary

Routes sharing a groupId do not share a counter with either built-in store. Each route gets the full max to itself, so a group of N routes allows N × max per window instead of max.

This is the exact problem #380 added the feature to solve — its motivation says "if we use per-route rate limit some clever users start to use all similar endpoints to achieve their task and 50 rps becomes 300 rps". With groupId set, 50 still becomes 300.

Steps to reproduce

This is the README's own /otp example, unchanged apart from counting the responses.

import Fastify from 'fastify'
import rateLimit from '@fastify/rate-limit'

const app = Fastify()
await app.register(rateLimit, { max: 100, timeWindow: '1 minute' })

const config = { rateLimit: { max: 3, timeWindow: '1 minute', groupId: 'OTP' } }
app.get('/otp/send',   { config }, async () => 'sent')
app.get('/otp/resend', { config }, async () => 'resent')

for (const url of ['/otp/send','/otp/send','/otp/send',
                   '/otp/resend','/otp/resend','/otp/resend','/otp/send']) {
  const r = await app.inject({ url })
  console.log(url.padEnd(12), r.statusCode, 'remaining=' + r.headers['x-ratelimit-remaining'])
}

Expected behaviour

The group allows 3 requests in the window; the 4th is refused, whichever route it arrives at.

Actual behaviour

6 are allowed — 3 at each route, each with its own x-ratelimit-remaining countdown.

/otp/send    200 remaining=2
/otp/send    200 remaining=1
/otp/send    200 remaining=0
/otp/resend  200 remaining=2
/otp/resend  200 remaining=1
/otp/resend  200 remaining=0
/otp/send    429 remaining=0

Cause

groupId is appended to the key, but the store is per route, so two routes in one group land on the same key in different stores.

In the onRoute hook, a route carrying its own config.rateLimit gets pluginComponent.store.child(mergedRateLimitParams), and:

- LocalStore.prototype.child returns a new LocalStore(...) — a fresh LRU per route. Identical keys never meet.
- RedisStore.prototype.child namespaces by routeOptions.routeInfo.method + routeOptions.routeInfo.url, so the route is already in the key prefix before groupId is appended to the suffix.

So groupId can only split one route's counter into several. It can never join two routes together.

Why it wasn't caught

test/group-rate-limit.test.js has a case named "With multiple routes and custom groupId" — but it gives the two routes different group ids (group1, group2) and asserts they don't interfere. There is no test anywhere for two routes sharing one groupId, which is the only thing the feature is for.

It works if the store honours the group

A custom store that namespaces by routeOptions.groupId instead of by the route makes the README example behave as documented — 3 
allowed across both routes, 4th refused. So the parameter already reaches the store; the two built-in stores just don't use it. 

Suggested fix

In both built-in stores, prefer routeOptions.groupId over the route's identity when deriving the child's namespace, falling back to today's behaviour when no groupId is set:

- LocalStore.child: with a groupId present, return a store sharing the parent's LRU with the group as a key prefix, rather than a new LRU.
- RedisStore.child: with a groupId present, use it in the key prefix in place of method + url.

Plus a test with two routes sharing one groupId, asserting the group total.

Impact

Anyone who adopted groupId for the reason #380 describes still has the multiplication it was meant to remove, and their limit reads correct in review. Worth a note in the release when it is fixed, since users may have sized max around the current behaviour. 
  

  

### Link to code that reproduces the bug

https://github.com/fastify/fastify-rate-limit/blob/main/README.md

### Expected Behavior

The group allows 3 requests in the window; the 4th is refused, whichever route it arrives at.

arpitadhage:
Fixes #463

## Problem
`groupId` was appended to the rate-limit key, but `.child()` still scoped
storage per route:
- `LocalStore.child()` created a fresh LRU per route, so identical keys
  in different LRUs never collided.
- `RedisStore.child()` always namespaced by `method + url`, putting the
  route in the key prefix before `groupId` was appended to the suffix.

Result: a group of N routes allowed N × max per window instead of max,
defeating the exact abuse pattern groupId (#380) was built to prevent.

## Fix
- `LocalStore.child()`: when `routeOptions.groupId` is set, the child
  store now shares the parent's LRU cache instead of creating a new one,
  so routes in the same group collide on the same key.
- `RedisStore.child()`: when `routeOptions.groupId` is set, it's used as
  the key-prefix namespace in place of `method + url`.
- Both fall back to today's per-route behavior when no `groupId` is set.

## Testing
Repro from the issue (3 max, groupId 'OTP', alternating /otp/send and
/otp/resend):

Before:
    /otp/send    200 remaining=2
    /otp/send    200 remaining=1
    /otp/send    200 remaining=0
    /otp/resend  200 remaining=2   <- bug: fresh counter
    /otp/resend  200 remaining=1
    /otp/resend  200 remaining=0
    /otp/send    429 remaining=0

After:
    /otp/send    200 remaining=2
    /otp/send    200 remaining=1
    /otp/send    200 remaining=0
    /otp/resend  429 remaining=0   <- shared counter, correctly refused
    /otp/resend  429 remaining=0
    /otp/resend  429 remaining=0
    /otp/send    429 remaining=0

Added a regression test ("two routes sharing one groupId share the
counter") to test/group-rate-limit.test.js — the existing suite only
tested that *different* groupIds stay isolated, never that the *same*
groupId is actually shared, which is why this shipped broken.

Full local suite passes except three pre-existing Redis timing tests
(`Redis with continueExceeding should not always return the timeWindow
as ttl`, `When continue exceeding is off under route (Redis)`,
`Route-specific exponential backoff with redis store`), which I confirmed
fail identically on a clean, unmodified `main` in my Windows/WSL Redis
setup — unrelated to this change.

## Note for changelog
This is a behavior fix, not a new feature — anyone currently using
`groupId` has been getting N × max instead of max. After this fix their
effective limit will be stricter and correct. Existing users of the
`groupId` option may need to re-tune `max` if they had unknowingly sized
around the old, broken multiplication.


mcollina:
CI is red

arpitadhage:
Fixed the lint issue (missing newline at EOF). Could a maintainer please approve the CI run?

arpitadhage:
Squashed into a clean commit and fixed the EOF newline issue. Could you please approve the CI run? Thanks!

arpitadhage:
Fixed the branch coverage issue by adding a test for the RedisStore.child() fallback path (when no groupId is set). Could you please approve the CI run? Thanks!

Tony133:
CI is red

